PR #44, opened 2018-04-25
Finish chap1 matlab to python conversions

Duane321:


Duane321:
Cool - I've made those changes. A few notes:

- save_fig in the utils.util file serves as our saving function.
- We now have PolyFitRidgeLasso, but I had to make a few adjustments to get the plots to look how intend: the lambdas for ridge/lasso are different then what's in the textbook. I was going for a squiggly look vs a smooth look and working with sklearn necessitated this change. Also, I changed the seed of PolyDataMake with the thibaux option (though this was included in the save_fig commit by mistake).

More to come


- save_fig in the utils.util file serves as our saving function.
- We now have PolyFitRidgeLasso, but I had to make a few adjustments to get the plots to look how intend:

PR #53, opened 2019-03-29
svd image demo with clown picture saved in data. numpy's SVD always r…

andrewnc:
…eturns the singular values in order, and so the final plot looks different than the MATLAB version. I'm not sure the fi, but it is potentially something we should revisit


murphyk:
The problem is that you were only randomly permuting the rows, which does
not destroy the correlation structure (as you can see by plotting the
permuted image). I have checked in a fix.

PR #57, opened 2019-04-09
add the natural gradient demo, with save figures, 

andrewnc:
runs slow with 1000000 points but can be reduced to 1000 to run quickly. Next step will be to add some comments to clarify a lot of the math. Similarly, it should be noted that the G\dL step in python is solved using least squares which can have different shapes than the MatLab solver according to this https://stackoverflow.com/questions/33559946/numpy-vs-mldivide-matlab-operator. 

Additionally, the original MatLab file did not have any names for the output figures. As such, I included save_fig with seemingly reasonable names, but these can easily be removed if desired.

### DIFF
--- a/book/nat_grad_demo.py
+++ b/book/nat_grad_demo.py
@@ -1,0 +1,98 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy.linalg import sqrtm
+from pyprobml_utils import save_fig
+
+
+def L_dL_G(theta, X, sf):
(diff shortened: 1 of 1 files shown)


murphyk:
Thanks! But gradient descent shouldn't be that slow. Please consider using
https://github.com/google/jax
to make a faster version. (Even on a CPU their jit function can provide
significant speedups. See also vmap.)
(I will add a section to the book describing jax, since I plan to use it
for many future coding examples, but for now you should
just read their documentation. This will be worth your time, since jax is
an incredible power tool that is worth learning :)

andrewnc:
ok great! I'll look into that!

murphyk:
Also the generated pdf files are massive (6MB!) Switch to png?

andrewnc:
That's likely a good idea. I think it has to do with the fact that we are using ~1000000 steps. When run using 1000 steps the form of the graphs are largely similar. The Matlab code was similarly slow. I'm excited to see how Jax can help.

PR #64, opened 2019-05-30
Converted shrinkcovdemo.m to Python.

jdf22:
Please extract the covcond function to a library script if required for other book scripts.


jdf22:
The shrinkcov.py file contains just one function. All it does is call sklearn.covariance.LedoitWolf and return the unbiased estimate in addition to that estimator's return values. You might want to scrap this library function and just use sklearn.covariance.LedoitWolf directly in place of the converted shrinkcov(X).

PR #66, opened 2019-09-15
add pcaDemo2d for fig 2.70

Duane321:
Hey, this is a PR for figure 2.70. My approach is to just recreate the image as I see it in the book, rather than tracing through the matlab code and recreate that directly. Happy to hear any feedback.


murphyk:
Thanks! Next time please save as pdf file :)

PR #82, opened 2021-01-12
add parzen-window2 example

gerdm:
* Add Python script that replicates figure 14.7 (PML 1st edition / parzenWindow2)

The following image is the result of the script
parzen-window


mjsML:
Thanks @gerdm

PR #92, opened 2021-02-14
Update mix_gauss_demo_faithful.py

kzymgch:
The return value "results" is not assigned.


mjsML:
Thank you @kzymgch 👍🏼

PR #97, opened 2021-03-02
Update gp_classify_spaceflu_1d_pymc3.py

kzymgch:
Currently there is no "vars" and it has been replaced by "var_names".


mjsML:
@kzymgch thank you for spotting this.

@murphyk an example of a new lib version that breaks working code to our conversation earlier.

PR #98, opened 2021-03-03
Update dtree_sensitivity.py

kzymgch:
Call 'fit' with appropriate arguments before using this estimator.
This is not PR, but The last image "dtree_iris_depth2_ensemble.pdf" can't be saved because "fit" function is not implemented in "prefit_voting_classifier.py".
and the link of this script is broken.

### DIFF
--- a/scripts/dtree_sensitivity.py
+++ b/scripts/dtree_sensitivity.py
@@ -117,6 +117,7 @@ if 0:
     eclf = VotingClassifier(
         estimators=[('orig', tree_clf), ('tweaked', tree_clf_tweaked)],
         voting='hard')
+    eclf.fit(X, y)
     plot_surface(eclf, X, y, xnames, ynames)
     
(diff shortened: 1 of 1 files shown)


mjsML:
@murphyk IIUC if we set the voting to hard this will throw an exception? ... thoughts?

murphyk:
The point of the example is to ensemble 2 previous fit trees.
I don't want to refit them!
The code is from here: https://gist.github.com/tomquisel/a421235422fdf6b51ec2ccc5e3dee1b4
It runs for me on my laptop. But you are right that the link from the book is broken.
It should point to https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/figures/chapter18_figures.ipynb#scrollTo=jDKa3vK_SPBq.
I will fix that.

kzymgch:
Thank you for your comment.
When I tried it again, there was no problem even if I didn't refit.

PR #285, opened 2021-03-15
Converted kpcaScholkopf to python 

ashishpapanai:
**figure number 20.35 in the book** 
#### Incorrect Figure, Trying to fix it. 
kpca


ashishpapanai:
@mjsML, What could be the reason for the absurd shapes of contours?

PR #286, opened 2021-03-15
Figure 18.1

Abdelrahman350:
Implementing regtreeSurfaceDemo which creates Figure 18.1

### DIFF
--- a/scripts/regtreeSurfaceDemo.py
+++ b/scripts/regtreeSurfaceDemo.py
@@ -1,0 +1,56 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from mpl_toolkits.mplot3d import axes3d
+
+x11 = 5
+x12 = 3
+x21 = 7
+x22 = 3
+
+r = np.arange(2, 12, 2)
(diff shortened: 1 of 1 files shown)


mjsML (changes requested):
18.1a is missing, 18.1b doesn't match 100% (if you look in the version of the book is slightly different).

Abdelrahman350:
> 18.1a is missing, 18.1b doesn't match 100% (if you look in the version of the book is slightly different).

18.1a is not plotted by Matlab, rather adapted from a reference that's why I didn't plot it using Python. Figure 18.1b is different in colors only and I will try to enhance it further.

mjsML:
> > 18.1a is missing, 18.1b doesn't match 100% (if you look in the version of the book is slightly different).
> 
> 18.1a is not plotted by Matlab, it's adapted from a reference that's why I didn't plot it using Python. Figure 18.1b is different in colors only and I will try to enhance it further.

I see, for 18.1b look at the axis, there is a gap between the hyperplane and the axis in your version vs. the book version.

PR #289, opened 2021-03-17
converted kpcaScholkopf to python

always-newbie161:
This is the plot generated by the code.

Screenshot 2021-03-17 at 4 30 16 PM


murphyk (changes requested):
Please replace the doubly nested for loops when computing the Gram matrix with the more efficient
sklearn.metrics.pairwise.euclidean_distances

murphyk:
This function is already written: https://github.com/probml/pyprobml/blob/master/scripts/kpcaScholkopf.py

PR #295, opened 2021-03-18
Converted logregLaplaceGirolamiDemo to python 

### LINKED ISSUE #201: Convert logregLaplaceGirolamiDemo to python
This is figure number 10.15 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/logregLaplaceGirolamiDemo.m)

ashishpapanai:
### Generates: figure number 10.14 & 10.15
Closes #201

logregLaplaceGirolamiData
logregLaplaceGirolamiPrior
logregLaplaceGirolamiNLL
logregLaplaceGirolamiPost

### DIFF
--- a/scripts/logregLaplaceGirolamiDemo.py
+++ b/scripts/logregLaplaceGirolamiDemo.py
@@ -1,0 +1,61 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy.stats import multivariate_normal
+import os
+
+if os.path.isdir('scripts'):
+    os.chdir('scripts')
+np.random.seed(0)
+fs = 14
+N = 30
+D = 2
+mu1 = np.concatenate((np.ones((N, 1)), 5*np.ones((N, 1))), axis=1)
+mu2 = np.concatenate((-5*np.ones((N, 1)), 1*np.ones((N, 1))), axis=1)
+class1_std = 1
+class2_std = 1.1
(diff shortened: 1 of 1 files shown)


ashishpapanai:
@murphyk @mjsML, The following lines use files present in MATLAB tools, 
```
pp =  preprocessorCreate('addOnes', false, 'standardizeX', false);
model = logregFitBayes(X, t, 'method', 'laplace', 'lambda', 1/alpha, 'preproc', pp);
```
I am unable to find a python equivalent for the same, and converting each MATLAB file in a single file will make the code way too big and it won't be reusable as well. Should I work on converting the files in MATLAB Tools to python or Could you please suggest any python equivalent if you know any?

murphyk:
The preprocessorCreate is just like skearn pipeline. In this case, the matlab code does not preprocessing,
so you can ignore it. As for logregFitBayes, you can find a Python version here:
https://bayes-logistic.readthedocs.io/en/latest/usage.html
(I have not tried it, not sure if it works.)

murphyk:
Please avoid hard-coding 'magic numbers; like 25291 in the line below
```
Log_Prior = Log_Prior.reshape((25921, 1))
```
I

ashishpapanai:
Tried, It doesn't solve the purpose of getting 
```
wMAP = model.wN;
C = model.VN; 
```

murphyk:
Ok, well you'll just have to write the code from first principles (converting math to numpy :)

PR #296, opened 2021-03-18
Added implementation for kpcaScholkopf

### LINKED ISSUE #232: Convert kpcaScholkopf to python
This is figure number 20.35 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/kpcaScholkopf.m)

animesh-007:
Added python implement for kpcaScholkopf.py
Fixes #232 
Screen Shot 2021-03-18 at 10 03 05 PM

### DIFF
--- a/scripts/kpcaScholkopf.py
+++ b/scripts/kpcaScholkopf.py
@@ -4,6 +4,7 @@
 import numpy as np
 import matplotlib.pyplot as plt
 import scipy.linalg as la
+from scipy.spatial.distance import pdist, cdist, squareform
 
 rbf_var = 0.1
 xnum = 4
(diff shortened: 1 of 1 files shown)


animesh-007:
@mjsML @murphyk I have a small doubt. In this problem, eigenvalues are dependent on the random number generator because of which I'm not able to get same eigenvalues. Let me know if this is okay, if not I will  copy matlab random number 
array in my code or you could suggest me something better.

murphyk:
Don't worry about the random seed, the results look fine.
However please eliminate the double for loop (for i, for j) and use 
scipy.spatial.distance.pdist to compute the Gram matrix.

animesh-007:
> Don't worry about the random seed, the results look fine.
> However please eliminate the double for loop (for i, for j) and use
> scipy.spatial.distance.pdist to compute the Gram matrix.

Made suggested changes.

mjsML (approved):
Looks good to me.

Issue #189, opened 2021-03-11
Convert gaussEvec to python


mjsML:
This is figure number 7.6 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/gaussEvec.m)

shivaditya-meduri:
https://github.com/probml/pyprobml/pull/297
Please find my pull request for this issue, Thank you.

PR #297, opened 2021-03-19
Convert gaussEvec to python (Issue-189)

shivaditya-meduri:
Python implementation of gaussEvec.m for Figure 7.6.
Figure created by python file shown below:

Figure_1

### DIFF
--- a/scripts/gaussEvec.py
+++ b/scripts/gaussEvec.py
@@ -30,13 +30,13 @@ def main():
               transform=rot + base)
     plt.arrow(0, -0.5, 0, 1, length_includes_head=True, width=0.015, head_width=0.05, head_length=0.05, color='blue',
               transform=rot + base)
-    plt.arrow(-1.2, 0.05, 0, 0.5, length_includes_head=True, width=0.015, head_width=0.05, head_length=0.05,
+    plt.arrow(-1.2, 0, 0, 0.5, length_includes_head=True, width=0.015, head_width=0.05, head_length=0.05,
               color='black', transform=rot + base)
-    plt.arrow(0.05, -0.5, 1.2, 0, length_includes_head=True, width=0.015, head_width=0.05, head_length=0.05,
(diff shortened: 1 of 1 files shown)


Koenie1980 (approved):

animesh-007:
Fixes #189

murphyk:
In colab, I get the figure below, which looks rather ugly....
Screen Shot 2021-03-19 at 5 14 06 PM

shivaditya-meduri:
Hi @murphyk, Actually the plot isn't wrong, The image I uploaded above is the generated plot, I think the problem is with the dimensions of the plot being different from the Google collab's dimensions of 640x480 which is causing the plot to be cropped at the edges. I will change the dimensions of the plot so that it can display properly on python notebooks as well.

mjsML (changes requested):
The figure doesn't display properly in Google's Colab (which is the primary runtime environment that is used). Please add logic to make the figure as portable as possible (across resolutions and platforms).

shivaditya-meduri:
image

This image is generated using the code in Google collab

@murphyk , Please review the changes that I've made, I fixed the scaling issue and it should work well on all platforms.

shivaditya-meduri:
Thank you for your patience!

shivaditya-meduri:
> The figure doesn't display properly in Google's Colab (which is the primary runtime environment that is used). Please add logic to make the figure as portable as possible (across resolutions and platforms).

Hi @mjsML , Please find the updated code and let me know if there are still problems. I appreciate your time.

mjsML (approved):

mjsML:
@shivaditya-meduri should the black arrows touch the blue arrows as the original figure?

shivaditya-meduri:
Dear @mjsML , I changed the double arrow to touch the blue line exactly. Thank you for your time

shivaditya-meduri:
image

@mjsML , please let me know if this solves the problem!

Issue #185, opened 2021-03-11
Convert neymanPearson2 to python


mjsML:
This is figure number 5.15 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/neymanPearson2.m)

Garvit9000c:
Started Working on it.

Garvit9000c:
Converted Successfully, part(a) 5.15 , Pull request initiated.

PR #302, opened 2021-03-20
Create neymanPearson2.py

Garvit9000c:
index

The Only Difference is , alpha and beta region are shaded continuously instead of Multiple consecutive Lines.


mjsML:
Please mention issue number like closes #185

mjsML (approved):
use matpplotlib to render labels properly as per their [documentation](https://matplotlib.org/stable/tutorials/text/mathtext.html)

Garvit9000c:
Ohk Sir, 
Now on will remember these two points.
*Linking the issue with pull request
*Using proper matplotlib annotation to render mathematical expressions.

murphyk (changes requested):
There is no need to use jax for such a simple example. Please rewrite to use vanilla numpy/scipy.

Garvit9000c:
OK Sir

Issue #212, opened 2021-03-11
Convert linregRbfDemo to python


mjsML:
This is figure number 13.24 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/linregRbfDemo.m)

nappaillav:
![image](https://user-images.githubusercontent.com/43855961/111992512-8542a200-8b3b-11eb-84cf-404a30f92ed6.png)
This is the figure that i'm getting @mjsML, the `KernelRidge(alpha=l, kernel='rbf')` works as expected  as `linregFit ` in this implementation. An help here could be very useful.

mjsML:
@nappaillav I believe you'd need to revise the data generation portion. Also the right column is squished for some reason.

nappaillav:
@murphyk @mjsML As per my understanding the prediction is not right. 
I'm trying to see the right replacement for `lineregfit` from the matlab code. Currently i'm using KernelRidgeregression function from sklearn. 
The data i.e generation is very similar to the one in the matlab figure 13.24

mjsML:
@nappaillav I see 19 points on your graph while the matlab one has 21 :) (maybe the dots need to be bigger as per the original figure?) Also the distribution doesn't look quite the same (I have not looked at your code I'm just looking visually). 
The [example](https://scikit-learn.org/stable/modules/kernel_ridge.html) on sklearn cites @murphyk 1st edition 🙂 , I believe it should work 🙂

Issue #99, opened 2021-03-03
convert legacy matlab code to python


murphyk:
If you want to contribute a Python version of a figure, follow these steps:

- Check the list of open issues for ones with a name like "Convert foo to Python". Such issues have the tag 'Figures'.
Each issue has a link to the Matlab file that was used to generate the  figure, and a figure number.
Figure numbers refer to this version of the book: https://github.com/probml/pml-book/releases/tag/2021-03-08. 
Consult the pdf file to see what the figure should look like.
  
- Follow [these guidelines](https://github.com/probml/pyprobml/blob/master/CONTRIBUTING.md) for how to contribute code.

- Test your code works in Google colab, and make a pull request. Please include the generated image(s) in your PR, and add a tag which mentions the issue you are solving.

- [Here](https://github.com/probml/pyprobml/pulls?q=is%3Apr+author%3Agerdm) is an example of good PR.

- If you are a student and want to be considered for the **Google Summer of Code**, please apply [here](https://summerofcode.withgoogle.com/get-started/) after **March 29th**. You should have at least 2 PRs succesfully added to this repo before applying. Please follow the style guidelines mentioned above.

ryzbaka:
Hi @murphyk , I'm interested in converting these examples to python with JAX. 

Is there anything I should know before getting started?

Garvit9000c:
Hi @murphyk Sir, I am Highly Enthusiastic , in converting These example to python with JAX.
And i am Already Building a Project with JAX , will upload the repo soon!
 What should be my First Step in Getting started!

murphyk:
I will provide more details on how to contribute soon.

Garvit9000c:
Ok Sir, That's crystal clear now!🙌

animesh-007:
@murphyk I would like to contribute to the project of converting Tensorflow DL examples to JAX / Flax / Haiku.
Could you guide me on this?

ryzbaka:
I downloaded the version of the book from the link [here](https://github.com/probml/pml-book/releases/tag/2021-03-08) as mentioned in the issue description.
Figure number 3.13 in the mcodes.txt file is a demo of Gaussian imputation, whereas in the book figure 3.13 is a visualization of the first 25 digits of MNIST.

![image](https://user-images.githubusercontent.com/26931277/110791147-e0e97180-8297-11eb-8989-4db08b3da9a4.png)

It'd be great if someone could help me out with this issue.

mjsML:
[...]

PR #306, opened 2021-03-22
Linear RBF demo

nappaillav:
Incomplete as Kernel regression is not working 
Issue #212 

This is the graph I'm getting 

image


mjsML:
closed as per the [guideline](https://github.com/probml/pyprobml/issues/99#issuecomment-803311745)

PR #311, opened 2021-03-24
Converted kernelRegressionDemo to python

### LINKED ISSUE #215: Convert kernelRegressionDemo to python
This is figure number 16.10 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/kernelRegressionDemo.m)

always-newbie161:
Figure:
Screenshot 2021-03-25 at 12 55 55 AM

Closes #215
@murphyk, @mjsML

### DIFF
--- a/scripts/kernelRegressionDemo.py
+++ b/scripts/kernelRegressionDemo.py
@@ -1,0 +1,25 @@
+from skfda import preprocessing as pp
+from skfda import FDataGrid
+
+np.random.seed(0)
+N = 100
+x = 10*(np.linspace(-1,1,100).reshape(-1,1))
+ytrue = np.array([math.sin(abs(el))/(abs(el)) for el in x]).reshape(-1,1)
+noise = 0.1
+y	= ytrue + noise*np.random.randn(N,1)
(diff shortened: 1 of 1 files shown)


mjsML:
updated colors as per [updated guidelines](https://github.com/probml/pyprobml/blob/master/CONTRIBUTING.md) image

mjsML (approved):
Looks good, however is the the [skfda](https://github.com/GAA-UAM/scikit-fda) dependency required?

always-newbie161:
> Looks good, however is the the [skfda](https://github.com/GAA-UAM/scikit-fda) dependency required?

Yes actually (can be installed using "!pip install scikit-fda"), the smoother is present in this package.

murphyk:
This works, but scikit-fda is a big package and slow to install. Is there no light weight solution?

always-newbie161:
> This works, but scikit-fda is a big package and slow to install. Is there no light weight solution?

As I found this as a spot-on function in this package for this problem, I used it to avoid any errors, If needed I would implement it myself using online resources.

PR #316, opened 2021-03-27
gprDemoNoiseFree

### LINKED ISSUE #219: Convert gprDemoNoiseFree to python
This is figure number 17.8 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/gprDemoNoiseFree.m)

nappaillav:
Closes #219

Figure 1) 
image

Figure 2)
image

Figure 3)
image

@mjsML @murphyk

### DIFF
--- a/scripts/gprDemoNoiseFree.py
+++ b/scripts/gprDemoNoiseFree.py
@@ -1,6 +1,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from numpy.linalg import cholesky
+from scipy.spatial.distance import cdist
 
 
 def gaussSample(mu, sigma, n):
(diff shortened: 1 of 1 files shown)


mjsML (commented):
It seems that the figure is different from the one in the book? I haven't looked at the code yet.

nappaillav:
@mjsML I believe that the difference is because of the GaussSample function, because of which there is randomness in the plot.

murphyk:
Please use scipy.spatial.distance.cdist to avoid for loop when computing sq_dist kernel.

murphyk (changes requested):
Please use scipy.spatial.distance.cdist to avoid for loop when computing sq_dist kernel.

nappaillav:
@murphyk updated with cdist

PR #323, opened 2021-03-29
mlpPriorsDemo2_jax

Abdelrahman350:
My implementation for mlpPriorsDemo2 using jax.

### DIFF
--- a/scripts/mlpPriorsDemo2_jax.py
+++ b/scripts/mlpPriorsDemo2_jax.py
@@ -1,0 +1,111 @@
+# Load JAX
+import jax
+import jax.numpy as jnp
+import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


animesh-007:
Hi @Abdelrahman350. Could you attach screenshots for the figures generated by your code and also mention the issue number which this PR closes? 😇
You could also read the guidelines provided by one of the mentors in case if you missed them.
[Link to guidelines](https://github.com/probml/pyprobml/issues/99#issuecomment-803311745)

murphyk:
What I had in mind was to use a pytree of parameters, to enable any number of layers, and to avoid explicitly having W1, b2, W2, b2 etc. But it's not that important, so I'll just accept this one as is.

PR #324, opened 2021-03-29
Converted svmCgammaDemo to python

### LINKED ISSUE #222: Convert svmCgammaDemo to python
This is figure number 17.19 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/svmCgammaDemo.m)

always-newbie161:
Closes #222
@mjsML @murphyk 

(PS: 'hastieMixture.mat' seem to be unavailable in the repo, So I used it from "probml" while testing"
  if it is OK I would add it in "/pyprobml/data/" and merge with this commit.)

I am not sure if CV.mean is used as the CV_estimate( I thought It would be standard error of CV.mean), But the matlab code seems to use CV.mean and I went with the same, these are the plots.

(I didn't find all the reference figures but probably mine seems to be different from them.)

Any suggestions would help me do them better :)

Screenshot 2021-03-29 at 10 53 31 PM

Screenshot 2021-03-29 at 10 53 20 PM

Screenshot 2021-03-29 at 10 53 10 PM

Screenshot 2021-03-29 at 10 52 50 PM

Screenshot 2021-03-29 at 10 44 37 PM
Screenshot 2021-03-29 at 10 44 50 PM
Screenshot 2021-03-29 at 10 44 59 PM

### DIFF
--- a/scripts/svmCgammaDemo.py
+++ b/scripts/svmCgammaDemo.py
@@ -1,0 +1,92 @@
+import numpy as np
+from scipy.io import loadmat
+from sklearn.svm import SVC
+from sklearn.model_selection import cross_val_score
+from sklearn.metrics import make_scorer, zero_one_loss
+import matplotlib.pyplot as plt
+from cycler import cycler
+
+# Using Colorblind friendly colors
+
+CB_color = ['#377eb8', '#ff7f00']
+
+cb_cycler = (cycler(linestyle=['-', '--', '-.']) * cycler(color=CB_color))
+plt.rc('axes', prop_cycle=cb_cycler)
+
+# -------------------------------------------
+# loading data
+
+data = loadmat('hastieMixture.mat')
+X = data['ans'][0][0][0]
+y = data['ans'][0][0][1].astype('int32')
+bayeserror = data['ans'][0][0][8]
+perm = np.random.permutation(X.shape[0])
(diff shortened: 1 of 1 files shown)


murphyk:
Please add 'hastieMixture.mat'  in a separate PR.

murphyk (commented):
Next time, please do the following:
- set np.random.seed(0)
- use loadmat('../data/hastieMixture.mat')
- save pdf version of figures to ../figures/foo.pdf, where foo matches the names used in the matlab script. This will let me swap out old figures for new in the book.

In this case, I'll just make these changes for you.

always-newbie161:
> Next time, please do the following:
> 
> * set np.random.seed(0)
> * use loadmat('../data/hastieMixture.mat')
> * save pdf version of figures to ../figures/foo.pdf, where foo matches the names used in the matlab script. This will let me swap out old figures for new in the book.
> 
> In this case, I'll just make these changes for you.

@murphyk  Sure, I will definitely do that next time, but for some data files "loadmat" from scipy.io is not working, this is because of newer version of some files(Eg:"bishop2class.mat")

murphyk:
Here's the tweaked version of your file, FYI.
https://github.com/probml/pyprobml/blob/master/scripts/svmCgammaDemo.py

PR #329, opened 2021-04-02
Converted probitRegDemo to python

### LINKED ISSUE #208: Convert probitRegDemo to python
This is figure number 12.2 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/probitRegDemo.m)

always-newbie161:
@murphyk @mjsML 

Closes #208

I tried to implement Probit Regression using EM and Minfunc(uses bfgs(or quasi-newton))  for log-likelihood estimation. I used the book and the matlab code as a guide. 

But I think LL is not converging properly😅(probably in minfunc). I request you to review this code and give suggestions 

Screenshot 2021-04-02 at 11 38 05 AM

Screenshot 2021-04-02 at 11 37 50 AM

### DIFF
--- a/scripts/probitRegDemo.py
+++ b/scripts/probitRegDemo.py
@@ -1,0 +1,134 @@
+import numpy as np
+from scipy.stats import norm
+from scipy.optimize import minimize, fmin_bfgs
+from matplotlib import pyplot as plt
+from cycler import cycler
+import pyprobml_utils as pml
+
+CB_color = ['#377eb8', '#ff7f00']
+
+cb_cycler = (cycler(linestyle=['-', '--', '-.']) * cycler(color=CB_color))
+plt.rc('axes', prop_cycle=cb_cycler)
+
+np.random.seed(0)
+
+value = 0
+
+class ProbitReg:
+
+    def __init__(self):
+        self.loglikhist = []
+        self.max_iter = 100
+        self.tolerance = 1e-4
+        self.w = []
+
+    def Probitloss(self, X, y, w):
+
+        value = 1 - norm.cdf(X @ w)
+        return -np.sum(y * norm.logcdf(X @ w)) - np.sum((1 - y) * np.ma.log(1 - (X@w)).filled(-(1e+20)))
+
+    def objfn(self, X, y, w, lambd):
+        return np.sum(lambd * np.square(w[1:])) - self.Probitloss(X, y, w)
+
+    def probRegFit_EM(self, X, y, lambd):
+
+        self.w = np.linalg.lstsq(X + np.random.rand(X.shape[0], X.shape[1]), y, rcond=None)[0].reshape(-1, 1)
+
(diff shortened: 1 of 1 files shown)


murphyk (changes requested):
Line 1. I would add references to the matlab source files (which are a bit hard to find!)
https://github.com/probml/pmtk3/blob/master/toolbox/SupervisedModels/probitReg/sub/probitRegFitEm.m
https://github.com/probml/pmtk3/blob/master/toolbox/SupervisedModels/probitReg/sub/probitRegFitMinfunc.m

Line 27. `value` seems ununsed.

Line 28. This is the cross entropy y*log(q) + (1-y)*log(1-q)?
Your math for the second term looks wrong. Also, what is np.ma.log?
Also, be careful, my matlab code assumed labels were -1,+1 ("plus-minus1"), instead of 0,1. You don't need to slavishly copy me, you should use 0,1.

Line 64. Don't cal it minfunc - that's the name of the particular matlab optimization package. The method should be called fit_gradient or something.

Line 77. minimize is a gradient free optimizer (zero'th order) - you should use BFGS, and provide the gradient and Hessian as args. I give explicit expressions in the book, but you can also use jax's autodiff to derive them (look at [this colab](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/logreg_jax.ipynb) for inspiration - it fits logistic regression using BFGS, where the gradient/Hessian is computed both by hand and using JAX.)

Line 117. Please compare your answers to https://www.statsmodels.org/stable/generated/statsmodels.discrete.discrete_model.Probit.html. The objective is convex, so all 3 methods should give the same results.

always-newbie161:
@murphyk 

Found my mistakes 

these are the figures
Screenshot 2021-04-04 at 4 33 57 PM

murphyk (changes requested):
This looks good. However, it seems [Probit.fit_regularized](https://www.statsmodels.org/stable/generated/statsmodels.discrete.discrete_model.Probit.fit_regularized.html#statsmodels.discrete.discrete_model.Probit.fit_regularized) uses L1 regularization, which cannot be directly compared to the L2 regularizer used by BFGS. Can you try [Probit.fit](https://www.statsmodels.org/stable/generated/statsmodels.discrete.discrete_model.Probit.fit.html#statsmodels.discrete.discrete_model.Probit.fit) instead, which comptues MLE? (For BFGS, set lambda to be small to approximate MLE)

Also, you define an estep function, but there is no mstep function. I realize the M step is just ridgeReg, but it might help clarity to define mstep() anyway.

Finally, please use [PEP 8 naming conventions](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

always-newbie161:
@murphyk 

- Used Probit.fit() with bfgs
- Added separate M step function and decreases the lam to 1e-2
- Used PEP 8 naming

PR #331, opened 2021-04-04

### LINKED ISSUE #174: Convert mixBerMnistEM to python

shivaditya-meduri:
mixBernoulliMnist

Closes #174 

### DIFF
--- a/scripts/gaussImputationDemo.py
+++ b/scripts/gaussImputationDemo.py
@@ -1,0 +1,125 @@
+# Author: Meduri Venkata Shivaditya
+# Illustration of data imputation using an MVN.
+import numpy as np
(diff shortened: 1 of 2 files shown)


shivaditya-meduri:
@murphyk and @mjsML, please review this PR.

mjsML (approved):

shivaditya-meduri:
Please ignore the 2nd commit, it is associated with another issue

mjsML:
@shivaditya-meduri , backup your work for both PRs, then, create a new branch **per issue** then create new PRs based on two branches, now this PR has two commits to two separate files.

shivaditya-meduri:
Thank you for your patience @mjsML

Issue #197, opened 2021-03-11
Convert fisherLDAdemo to python


mjsML:
This is figure number 9.5 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/fisherLDAdemo.m)

Drishttii:
Hi, I am Drishti an outreachy applicant. I would like to work on this issue.

mjsML:
> Hi, I am Drishti an outreachy applicant. I would like to work on this issue.

Please go ahead :)

PR #332, opened 2021-04-05
Convert fisherLDAdemo to python

Drishttii:
@murphyk and @mjsML,  please review the code.

#197 

Figures generated:

a) Fig. 9.4

FischerLDA_and_PCA_vectors

b) Fig. 9.5 (a)

Projection_of_points_on_fisher_vec

c) Fig. 9.5 (b)

Projection_of_points_on_pca_vec

Also, shall I delete the .pynb file?


mjsML (changes requested):
Please remove the ipynb file.
Your pull should contain the py file used to create the figures for now.

Drishttii:
Removed the .ipynb file. Thanks.

murphyk (changes requested):
I like that you implement both methods using sklearn and 'from scratch'. It would be even better if you check automatically that they give the same results. Also please refactor so the plotting code is shared across both methods. So the code should look something like this (where results is the relevant output):
```
results_flda_sklearn = flda_sklearn(data)
results_flda_manual = flda_manual(data)
assert np.allclose(results_flda_sklearn, results_flda_manual))
plot_results(results_flda_sklearn)

// repeat above using pca
```
 

```

Drishttii:
Thank you for your comments. I have made the suggested changes, please let me know if anything else is required.
@murphyk @mjsML

murphyk:
This is great. Next time please use `assert np.allclose(...)` instead of print statements :)

murphyk:
Also, if the matlab files is `fisherLDAdemo.m`, please call the python file `fisherLDAdemo.py`, since it is meant to be a drop-in replacement. (You named your file `convert_fisherldademo_to_python.py`!)

murphyk:
Also, please use `np.random.state(0)` for reproducibility.

Drishttii:
Thank you very much. I will keep these suggestions in mind :)

PR #336, opened 2021-04-06
jax instead of numpy

nouranali:


mjsML (commented):
Doesn't seem like a wise change, I'm not sure how the underlying XLA backend of JAX allocates GPU/TPU memory. Also by definition, this would create a few CUDA/TPU IO/ Thread requests more than required (you either need a PyTorch or JAX arrays).
Mixing JAX and PyTorch is [not recommended](https://github.com/pytorch/pytorch/issues/17569)

nouranali:
> Doesn't seem like a wise change, I'm not sure how the underlying XLA backend of JAX allocates GPU/TPU memory. Also by definition, this would create a few CUDA/TPU IO/ Thread requests more than required (you either need a PyTorch or JAX arrays).
> Mixing JAX and PyTorch is [not recommended](https://github.com/pytorch/pytorch/issues/17569)

Ok I'll try to use all JAX arrays

mjsML:
> > Doesn't seem like a wise change, I'm not sure how the underlying XLA backend of JAX allocates GPU/TPU memory. Also by definition, this would create a few CUDA/TPU IO/ Thread requests more than required (you either need a PyTorch or JAX arrays).
> > Mixing JAX and PyTorch is [not recommended](https://github.com/pytorch/pytorch/issues/17569)
> 
> Ok I'll try to use all JAX arrays

Sorry I was not clear, Since the core example is in PyTorch and mixing JAX and PT is not recommended, I would recommend you create a new example that uses JAX only e.g. `cifar5-grid_jax.py` ... However since is a demo of a private alpha service [grid.ai](https://grid.ai), Also this service doesn't support JAX officially yet.
If you have access to grid.ai and would like to show us the result of making JAX work on their service I would pursue this further. Otherwise, I would try to find another example to convert.

PR #339, opened 2021-04-07
Converted gaussImputationDemo to python

### LINKED ISSUE #171: Convert gaussImputationDemo to python
This is figure number 3.7 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/gaussImputationDemo.m)

shivaditya-meduri:
Hinton_Truth
Hinton_ImputationWithTrueParams
Hinton_Observed

Closes #171 

I implemented gauss Imputation basing it on gaussImpute() and gaussCondition() matlab files in pmtk3.googlecode.com. 

To randomly create sigma, I have used the sci-kit learn library instead of implementing randpd because of the excessive time delay in creating a positive semidefinite matrix as the probability of creating one SPD using randpd in python is less than 1.

Please note that the MATLAB implementation has randomly generated a matrix of size 20x10 but in the book I observed that a matrix of 8x10 is provided, therefore I provided the inputs d=8 and n=10, but the dimensions of the matrix can be changed by altering n & d in the code.

### DIFF
--- a/scripts/gaussImputationDemo.py
+++ b/scripts/gaussImputationDemo.py
@@ -1,0 +1,125 @@
+# Author: Meduri Venkata Shivaditya
+# Illustration of data imputation using an MVN.
+import numpy as np
+import matplotlib.pyplot as plt
+from sklearn.datasets import make_spd_matrix
+
+def is_pos_def(x):
+    #Check if the matrix is positive definite
+    j = np.linalg.eigvals(x)
+    return np.all(j>0)
+
+def gaussSample(mu, sigma, n):
+    # Returns n samples (in the rows) from a multivariate Gaussian distribution
+
+    a = np.linalg.cholesky(sigma)
+    z = np.random.randn(len(mu), n)
+    k = np.dot(a, z)
+    return np.transpose(mu + k)
+
+def gaussCondition(m, s, v, vv):
+    # p(xh | xv = visValues)
+
+    d = len(m)
+    j = np.array(range(d))
+    h = np.setdiff1d(j, v)
+    if len(h)==0:
+        mugivh = np.array([])
+        sigivh = np.array([])
+    elif len(v) == 0:
+        mugivh = m
+        sigivh = s
+    else:
+        shh = np.zeros((len(h), len(h)))
+        for lr, uu in enumerate(h):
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
@mjsML, @murphyk 
Please review this PR

mjsML (approved):

mjsML:
I updated the colors, please update the screenshots based on the new colors :)

murphyk (changes requested):
- please use longer, more readable variable names, and follow [PEP8 conventions](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), even if this differs from the matlab.
- please vectorize your code! eg to extract the hidden entries of the Sigma matrix and assign to the Sh matrix, you can use [fancy indexing](https://jakevdp.github.io/PythonDataScienceHandbook/02.07-fancy-indexing.html) to do it in parallel.
- please rename hinton to `hinton_diagram` and submit this as a separate, standalone PR. 
-

murphyk:
Here is an example of how extract a block submatrix.
First make the matrix
```
import numpy as np
X = np.arange(12).reshape((3, 4))
print(X)
row = np.array([0,  2])
col = np.array([0, 2, 3])
```
Here is the slow for-loop version:
```
X2 = np.zeros([len(row), len(col)])
for itarget, isrc in enumerate(row):
  for jtarget, jsrc in enumerate(col):
    X2[itarget, jtarget]  = X[isrc, jsrc]
print(X2)
```
Here is the preferred vectorized version
```
ndx = np.ix_(row, col)
X2 = X[ndx]
print(X2.shape) # (2,3)
print(X2)
```

murphyk:
It seems that [hinton_diagram](https://github.com/probml/pyprobml/blob/master/scripts/hinton_diagram.py) is already in pyprobml, so you can just import that and use it, rather than duplicating code.

shivaditya-meduri:
Hi @murphyk , I observed that the hinton_diagram.py has a few deprecated methods in the code and I created a pull request that will fix the bug.
The color schematic of hinton_diagram.py file is restricted to black and white, and the images in the book are colored differently, 
can you please confirm if the colors of the positive and the negative elements of the matrix can be black and white?

PR #340, opened 2021-04-07

### LINKED ISSUE #174: Convert mixBerMnistEM to python

shivaditya-meduri:
mixBernoulliMnist
Closes #174

### DIFF
--- a/scripts/mixBerMnistEM.py
+++ b/scripts/mixBerMnistEM.py
@@ -1,0 +1,170 @@
+# Author: Meduri Venkata Shivaditya
+# Bernoulli mixture model for mnist digits
+import numpy as np
+import matplotlib.pyplot as plt
+from sklearn.datasets import fetch_openml
+
+
+def bernoulli(data, means):
+    '''To compute the probability of x for each bernouli distribution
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
@mjsML, Thank you for your suggestion, I've created a new pullrequest

mjsML (approved):

murphyk (changes requested):
- Please vectorize all your code where possible. For example, never write `for i=1:N`! (Obviously EM iterations are sequential, and it's probably ok to loop over k=1:K, but not over data points.)
- replace your `bernoulli` function with https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bernoulli.html
- factor out the code that does parameter initialization and image plotting into subfunctions, to simplify main() 
- I find fetch_openml to be slow for getting MNIST. You may find the pytorch.vision dataloader, or the tensorflow datasets version, faster (both libraries are built in to colab).

shivaditya-meduri:
I made a new PR with only 1 commit, Please find it here : https://github.com/probml/pyprobml/pull/344

PR #341, opened 2021-04-07
Converted polyfitRidgeModelSel to python

### LINKED ISSUE #175: Convert polyfitRidgeModelSel to python
This is figure number 4.7 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/polyfitRidgeModelSel.m)

always-newbie161:
Closes #175

@murphyk,  @mjsML
Please review.

I implemented Bayes linear Reg with Gaussian Prior and Variational Bayes but I am not able to implement Empirical 
Bayes as I am not sure about netlab procedure and I probably need some guidance.

(But the needed figures didn't use the any of the EB attributes, So I made this PR)

Figures:
Screenshot 2021-04-08 at 1 06 25 AM
Screenshot 2021-04-08 at 1 06 33 AM
Screenshot 2021-04-08 at 1 06 40 AM
Screenshot 2021-04-08 at 1 06 52 AM
Screenshot 2021-04-08 at 1 06 45 AM


murphyk (changes requested):
- it seems that in fig 4.7b, although I plot the VB estimate, I don't explain it in this part of the book (this is a legacy figure from an older version). So please refactor this script into 2 separate scripts: polyfitRidgeCV.py and polyfitRidgeBayes.py. The former just uses ridgr regression and CV, the latter contains the Bayesian code. 
- who is Andreas C.Kapourani (kapouranis.andreas@gmail.com)? Is his code public? Can you add a pointer to it, or (if it's pyprobml) just import it?
-

always-newbie161:
> * it seems that in fig 4.7b, although I plot the VB estimate, I don't explain it in this part of the book (this is a legacy figure from an older version). So please refactor this script into 2 separate scripts: polyfitRidgeCV.py and polyfitRidgeBayes.py. The former just uses ridge regression and CV, the latter contains the Bayesian code. The former code can generate 4.7a, and a simplified version of 4.7b that just shows the CV curve (with error bars), without the VB curve.
> * who is Andreas C.Kapourani ([kapouranis.andreas@gmail.com](mailto:kapouranis.andreas@gmail.com))? Is his code public? Can you add a pointer to it, or (if it's pyprobml) just import it?

@murphyk, Sure, I will refactor it into two scripts and for the VB code I referenced a blog from "Andreas C.Kapourani"  which is public and this is the [blog](https://rpubs.com/cakapourani/variational-bayes-lr).

always-newbie161:
@murphyk 

- made the two two scripts 
- added the reference for VB
- formatted using pep8 styles

murphyk:
It seems like the train/test curves are slightly different for the 2 experiments - in particular the test minimum are in different
locations. Is this an x-axis scaling issue?
(Is lambda in ridge identical to alpha in VB? Or is one scaled by the std deviation?)
Here is the ridge plot

Screen Shot 2021-04-15 at 9 02 56 PM


And here is the VB plot

Screen Shot 2021-04-15 at 9 06 17 PM

Ideally these plots would be the same.
Of course the optimal lambda estimated by CV may not be identical to the optimal alpha estimated by VB, but it should be close.

Finally, as for https://rpubs.com/cakapourani/variational-bayes-lr, did you translate his R code to python?

always-newbie161:
> It seems like the train/test curves are slightly different for the 2 experiments - in particular the test minimum are in different
> locations. Is this an x-axis scaling issue?
> (Is lambda in ridge identical to alpha in VB? Or is one scaled by the std deviation?)
> Here is the ridge plot
> 
> Screen Shot 2021-04-15 at 9 02 56 PM
> 
> And here is the VB plot
> 
> Screen Shot 2021-04-15 at 9 06 17 PM
> 
> Ideally these plots would be the same.
> Of course the optimal lambda estimated by CV may not be identical to the optimal alpha estimated by VB, but it should be close.
> 
> Finally, as for https://rpubs.com/cakapourani/variational-bayes-lr, did you translate his R code to python?

@murphyk I found my mistake,  linear_model.Ridge() by default assumes bias term in X, as we are not adding ones to X, the plots differed. Now i added 'fit_intercept=False' to the model, now the mse_values are approximately same
[For the VB code, yes I used his R code as reference]. **But the bayes plot is from the gaussian prior bayes**.

Ridge plot
Screenshot 2021-04-16 at 3 03 13 PM

Bayes Plot
Screenshot 2021-04-16 at 3 03 22 PM

murphyk:
Yes, VB will not give exactly same answer as MAP estimation with Gaussian prior - that is to be expected.
It's interesting that VB gives a nice well defined U-curve, making it easy to see the optimal lambda/alpha, but CV does not (the curve just flattens out). This trend was apparent in the original matlab figure, but it's even clearer now :)

murphyk:
Ideally this code and https://github.com/probml/pyprobml/blob/master/scripts/linreg_poly_ridge.py could be all be combined in a consistent way. In the meantime I ensured the train/test plots are the same across both scripts.

always-newbie161:
@murphyk thank you :)

PR #343, opened 2021-04-10
Converted gaussImputationDemo to python

### LINKED ISSUE #171: Convert gaussImputationDemo to python

shivaditya-meduri:
Hinton_Observed
Hinton_Truth
Hinton_ImputationWithTrueParams

Closes #171
I implemented gauss Imputation basing it on gaussImpute() and gaussCondition() matlab files in pmtk3.googlecode.com.

To randomly create sigma, I have used the sci-kit learn library instead of implementing randpd because of the excessive time delay in creating a positive semidefinite matrix as the probability of creating one SPD using randpd in python is less than 1.

Please note that the MATLAB implementation has randomly generated a matrix of size 20x10 but in the book I observed that a matrix of 8x10 is provided, therefore I provided the inputs d=8 and n=10, but the dimensions of the matrix can be changed by altering n & d in the code.

Updated the code to include the following changes:
1. use longer, more readable variable names, and follow PEP8 conventions, even if this differs from the matlab.
2. vectorize your code! eg to extract the hidden entries of the Sigma matrix and assign to the Sh matrix, you can use fancy indexing to do it in parallel.

### DIFF
--- a/scripts/gaussImputationDemo.py
+++ b/scripts/gaussImputationDemo.py
@@ -1,0 +1,118 @@
+# Author: Meduri Venkata Shivaditya
+# Illustration of data imputation using an MVN.
+import numpy as np
+import matplotlib.pyplot as plt
+from sklearn.datasets import make_spd_matrix
+
+def is_pos_def(x):
+    #Check if the matrix is positive definite
+    j = np.linalg.eigvals(x)
+    return np.all(j>0)
+
+def gauss_sample(mu, sigma, n):
+    # Returns n samples (in the rows) from a multivariate Gaussian distribution
+
+    a = np.linalg.cholesky(sigma)
+    z = np.random.randn(len(mu), n)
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
Please review this PR, @murphyk , @mjsML

I made all the changes suggested.
 
Hi @murphyk , I observed that the hinton_diagram.py has a few deprecated methods in the code and I created a pull request that will fix the bug.
The color schematic of hinton_diagram.py file is restricted to black and white, and the images in the book are colored differently,
can you please confirm if the colors of the positive and the negative elements of the matrix can be black and white?

If you want me to change hinton_diagram.py to accept colors of the hinton diagram as input too and make it more flexible, I could do that as well.

PR #344, opened 2021-04-11

### LINKED ISSUE #174: Convert mixBerMnistEM to python

shivaditya-meduri:
mixBernoulliMnist

Closes #174

Implemented the changes suggested:
1. Vectorized code and replaced iterating over the data points to vectorized versions
2. replaced Bernoulli function with https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bernoulli.html
3. factored out the code that does parameter initialization and image plotting into subfunctions, to simplify main()
4. changed the data source from open_ml to tensorflow.datasets

### DIFF
--- a/scripts/mixBerMnistEM.py
+++ b/scripts/mixBerMnistEM.py
@@ -1,0 +1,162 @@
+# Author: Meduri Venkata Shivaditya
+# Bernoulli mixture model for mnist digits
+import numpy as np
+import matplotlib.pyplot as plt
+import tensorflow as tf
+from scipy.stats import bernoulli as bern
+
+
+def bernoulli(data, means, K):
+    '''To compute the probability of x for each bernouli distribution
+    data = N X D matrix
+    means = K X D matrix
+    prob (result) = N X K matrix
+    '''
+    N = len(data)
+    D = len(data[0])
+    # compute prob(x/mean)
+    # prob[i, k] for ith data point, and kth cluster/mixture distribution
+    prob = np.zeros((N, K))
+    for k in range(K):
+            b = lambda row : np.prod(bern.pmf(row, means[k]))
+            prob[:, k] = np.apply_along_axis(b, 1, data)
+    return prob
+
+
+def respBernoulli(data, weights, means, K):
+    '''To compute responsibilities, or posterior probability p(z/x)
+    data = N X D matrix
+    weights = K dimensional vector
+    means = K X D matrix
+    prob or resp (result) = N X K matrix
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
Hi @murphyk , @mjsML , I made the suggested changes. Please review this PR

murphyk (changes requested):
- The demo is very slow, because full batch EM on 60k examples is a bad idea. In the matlab code, I randomly subsampled N=1000 examples (after shuffling the rows); please do the same,
- Please expand the name 'resp' to 'responsibilities', and use PEP8 naming, not camel case, so `respBernoulli` becomes `compute_responsibilitie`, `bernoulliMStep` becomes `m_step`, `llBernoulli` becomes `expected_log_lik`, etc.
- Rename `bernoulli` (which is ambiguous) to `bernoulli_mixture_pmf`.

shivaditya-meduri:
Hi @murphyk, Did you mean that the current python code attached with this PR is very slow and I need to look into it or it is sufficiently fast enough and I could change the N to 10k?

murphyk:
It should suffice to just set N=10k. (The code seems as  vectorized as it can be, without resorting to jax...)

shivaditya-meduri:
Thank you for your time reviewing this code, I will make the changes ASAP

shivaditya-meduri:
Hi @murphyk , I changed N to 10k and the execution time increased almost 8 fold, Initially it was around 2 mins(N=1000), now it takes around 15 mins(N=10k) for execution. I am currently trying JAX to implement a faster version of the mixture of Bernoulli model.
What would be the limit before the execution is considered too slow? Depending on that we can train the model to the said number of images.

murphyk:
I like all the demos to run in under ~2 mins :). Using N=1000 and max_iter  = 5 meets that goal :). As for JAX, this may help, but the biggest speedup will come from an online EM algorithm, rather than full batch. For this I would suggest using stochastic variational inference. I will cover this in vol 2, and make an issue :)

shivaditya-meduri:
Thank you for your pateince!

murphyk:
If you like to get a head-start, a JAX version of this SVI code (that uses TFP) would be great: https://brendanhasz.github.io/2019/06/12/tfp-gmm.html

shivaditya-meduri:
Thanks @murphyk, I will start working on it.

murphyk:
I opened https://github.com/probml/pyprobml/issues/348 to track progress on this.

Issue #220, opened 2021-03-11
Convert gprDemoChangeHparams to python


mjsML:
This is figure number 17.9 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/gprDemoChangeHparams.m)

ashishpapanai:
Working on it.

ashishpapanai:
@mjsML 
I am unable to find any Python counterpart for the following functions:
`covfunc = {'covSum', {'covSEiso','covNoise'}};`
And it is not a very good idea to convert the MATLAB code of these three functions to Python, is there any suitable library/ code in your or @murphyk's knowledge?

mjsML:
@ashishpapanai I'm not a matlab nor a GP expert, a small google search yielded [this](https://www.mathworks.com/matlabcentral/answers/10269-what-does-mean). and I believe those underlying functions should have equivalents in [scikit-learn ](https://scikit-learn.org/stable/modules/gaussian_process.html#gpr-examples) and even an attempt to mimic matlab convention [here](https://python.hotexamples.com/site/file?hash=0xddec81393c1b5ccdca64319b5fa99207b205917ef148449682585b2d39435b6a&fullName=MLO-master/model/regressors.py&project=CustomComputingGroup/MLO).
I hope this was helpful.

Drishttii:
Hello, @ashishpapanai , I have tried solving this issue (#345 ). Please let me know if you have any suggestions. Thank you.

murphyk:
Yes, #345 solved this.

PR #346, opened 2021-04-11
Convert_gprdemochangehparams_to_python

Drishttii:
@mjsML and @murphyk, please review the code. Thanks :)

#220 

Figures generated:

a) Fig. 17.9 (a)
Plot-0

b) Fig. 17.9 (b)
Plot-1

c) Fig. 17.9 (c)
Plot-2


murphyk (changes requested):
I'm confused why you use covSEiso+covNoise to generate the training data, but then use a different RBF kernel to fit the data. Can you not use the same kernel for both? Also why do you need sklearn GP and Gpy?

Drishttii:
Thank you for the review. The function covSEiso is basically generating datapoints from an RBF kernel with parameters l=1.0, sigma_f=1.0 and since we need noisy datapoints I used covNoise with sigma_y=0.1 (as mentioned in the book and the matlab code). So, covSEiso and covNoise are made just for the sake of doing them. Now I am generating data normally. Also, yes the sklearn imports were not required. Thanks again :)

murphyk:
This is much simpler now - thanks!

Drishttii:
Thank you very much :)

PR #347, opened 2021-04-15
Figure 2.15: display x values in x-axis

hieuza:
Also: add figure title "Beta distributions" and ylabel pdf.


hieuza:
The new plot:

image

Issue #218, opened 2021-03-11
Convert gpKernelPlot to python


mjsML:
These are figures 17.3 and 17.4 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/gpKernelPlot.m)

PR #351, opened 2021-04-17
Converted the gpKernelPlot code

Prahitha:
#218 
Hello @murphyk and @mjsML please review my code and let me know what you think. I wasn't sure if the plots had to be the exact same color, I'll change them if they must match (images attached below). Thanks!
periodic_kernel
cosine_kernel
matern_kernel_1
matern_kernel_5


mjsML (changes requested):
It seems that figure 17.3 is generated using the same file [gpKernelPlot.m](https://github.com/probml/pmtk3/blob/master/demos/gpKernelPlot.m), Can I ask you to kindly add 17.3A and 17.3B too?

mjsML (approved):

PR #352, opened 2021-04-18
Convert_gprdemomarglik_to_python

Drishttii:
@murphyk and @mjsML please review the code.

Figures generated:

a) Fig. 17.10 (a)
Contour

b) Fig. 17.10 (b)
Plot-0

c) Fig. 17.10 (c)
Plot-1

### DIFF
--- a/scripts/gprdemomarglik.py
+++ b/scripts/gprdemomarglik.py
@@ -5,13 +5,17 @@ import pyprobml_utils as pml
 import numpy as np
 import matplotlib.pyplot as plt 
 import GPy
-from numpy.linalg import cholesky, det, inv
+from numpy.linalg import cholesky, det, inv, lstsq
 from scipy.linalg import solve_triangular
+import scipy.spatial.distance as spdist
+from scipy import linalg
+from scipy.optimize import minimize
 
 # Dataset according to matlab code
 n = 7
 np.random.seed(20) 
 xs = 15*(np.random.uniform(low=0, high=1, size=n) - 0.5).reshape((-1, 1))
+sigma_y = 0.1
 def K(p, q):
   p = p.transpose()
   p = p.transpose()
@@ -19,12 +23,15 @@ def K(p, q):
   q = np.tile(q, len(p))
   r = 0.5*np.square(p-q)
   return np.exp(r)
-
-w = K(xs, xs) + 0.01*np.eye(n)
+w = K(xs, xs) + (sigma_y**2)*np.eye(n)
 w = w.conjugate()
-fs = np.linalg.cholesky(w).dot(np.random.randn(n, 1)).reshape((-1, 1))  
+fs = np.linalg.cholesky(w).dot(np.random.randn(n, 1)).reshape((-1, 1))
+
+#xs = np.array([-1.3089, 6.7612, 1.0553, -1.1734, -2.9339, 7.2530, -6.5843]).reshape(-1, 1)
+#fs = np.array([1.6218, 1.8558, 0.4102, 1.2526, -0.0133, 1.6380, 0.2189]).reshape(-1, 1)
+
 n=41
-x = np.linspace(0.1, 80, n)
+x = np.linspace(0.1, 150, n)
 y = np.linspace(0.03, 3, n)
 X, Y = np.meshgrid(x, y)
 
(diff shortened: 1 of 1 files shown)


murphyk (changes requested):
- The dataset generator should use the same code as in https://github.com/probml/pyprobml/blob/master/scripts/gprDemoChangeHparams.py. (Note that sigma_y = 0.1).
- Ideally the surface plot of MLL will have 2 clear local optima, as in the book (corresponding to different explanations of the data, namely low sigma_y but small length scale (wiggly), or high  sigma_y but large length scale (smooth)). 
- Yoy should  plot the estimated function at the 2 local optimal of the MLL. The values of these hyper--parameters should be determined by optimization (from different starting points), and not hard coded, since the value you plot are customized to the matlab data (which has a different random seed, etc)
-

murphyk:
To simplify reproduction of the original figure, I ran the [original matlab code](https://github.com/probml/pmtk3/blob/master/demos/gprDemoMarglik.m), and extracted the 7 values of (xi,yi) that are used as the dataset (for computing MLL surface plot). Given this data, everything else should be deterministic (modulo choice of optimizer you use :).
`
>> xs'
ans =
   -1.3089    6.7612    1.0553   -1.1734   -2.9339    7.2530   -6.5843
>> fs'
ans =
    1.6218    1.8558    0.4102    1.2526   -0.0133    1.6380    0.2189
`
HTH!

Drishttii:
Thank you very much for the comments. 

I ran the [matlab code](https://github.com/probml/pmtk3/blob/master/demos/gprDemoMarglik.m) (as it is) but I was getting similar plot of the log-marginal likelihood as the one from my python code. (I took l=1, sigma_y=0.1 and fixed sigma_f=1)

Matlab figure: 
Contour

My figure:
Contour

So, I have tried to reproduce the results keeping data as mentioned in the matlab code and also using the one (hard code) you suggested. I optimized log-liklihood function to get the best possible hyperparams (l and sigma_y). But should I try all 41*41 starting points (l and sigma_y) or only a few from these?

murphyk:
So there must be some discrepancy between how matlab interprets the hyper-parameters and how GPy does (perhaps it's a standard deviation vs variance vs 1/variance issue?). In matlab I get this surface plot:

Screen Shot 2021-04-21 at 10 00 31 PM

Using the same data (the xs, fs that you commented out) in your python code, I get this

Screen Shot 2021-04-21 at 10 02 05 PM

You first need to make the surface plots match before worrying about where to start the optimizer.

murphyk:
Note that in the matlab code, the kernel includes the observation noise, I'm not sure what Gpy does...
```
function [out1, out2, out3] = gpS00evidence(X, input, target, mask, test);
...
% C(x^p,x^q) = v^2 * exp[-(x^p - x^q)'*inv(P)*(x^p - x^q)/2]
%            + u^2 * delta_{p,q}
```

Drishttii:
Hello, I tried to get the log-likelihood in two ways, from scratch and using Gpy. I am getting similar values of Z and hence the same contour plot from both methods. Also, I am getting almost similar values of Z when I ran the Matlab code. I have printed out the values of Z and it would be great if the comparison of Z values with your Matlab code can be done. Thanks :)

scripts/gprdemomarglik.py
```diff
@@ -0,0 +1,140 @@
...
+  r = 0.5*np.square(p-q)
+  return np.exp(r)
+w = K(xs, xs) + (sigma_y**2)*np.eye(n)
+w = w.conjugate()
+fs = np.linalg.cholesky(w).dot(np.random.randn(n, 1)).reshape((-1, 1))
+
+#xs = np.array([-1.3089, 6.7612, 1.0553, -1.1734, -2.9339, 7.2530, -6.5843]).reshape(-1, 1)
+#fs = np.array([1.6218, 1.8558, 0.4102, 1.2526, -0.0133, 1.6380, 0.2189]).reshape(-1, 1)
+
+n=41
```
gerdm, line 31: @Drishttii, have you tried playing with this parameter? I tried to run your code in Colab, but it got stuck for some reason. 😕

Drishttii: @gerdm, I will try that and see what happens .... also let me know what error you are getting .... whenever you have time :) Thank you for the suggestion.

gerdm: For sure! I will try again and let you know.

gerdm:
@Drishttii, it seems that the problem was with Matplotlib. The log-likelihood becomes very small for values in the lower-right region. I was able to reproduce the graph that @murphyk obtained

image

Specifying the levels in Matplotlib solves the problem. For me, it was

```python
plt.contour(*np.exp(P), Z[0, 0], levels=levels)
```

image


I will PR code with changes.

gerdm:
I yet have to clean my code, but you can look at the raw (dirty code) [here](https://github.com/gerdm/misc/blob/master/2021-07/gpr_demo_marlik.ipynb)

murphyk:
solved by https://github.com/probml/pyprobml/blob/master/scripts/gpr_demo_marlik.py

PR #353, opened 2021-04-18
Convert_gaussimputationdemoem_to_python

Drishttii:
@mjsML and @murphyk please review the code. Thanks :)

Figures generated:

a) Fig. 8.28 (a)
Imputation with true params

b) Fig. 8.28 (b)
Imputation with EM params

### DIFF
--- a/scripts/gaussimputationdemoem.py
+++ b/scripts/gaussimputationdemoem.py
@@ -1,4 +1,5 @@
 import pyprobml_utils as pml
+
 import numpy as np
 import matplotlib.pyplot as plt
 from sklearn.datasets import make_spd_matrix
@@ -54,7 +55,7 @@ def gauss_impute(mu, sigma, x):
              x_imputed[i, h] = mu_hgv[rr]
      return x_imputed
 
-def impute_em(X, max_iter = 100, eps = 1e-04):
+def impute_em(X, max_iter = 50, eps = 1e-04):
 
     nr, nc = X.shape
     C = np.isnan(X) == False                    # Identifying nan locations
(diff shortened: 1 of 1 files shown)


Drishttii:
Issue #196

murphyk (changes requested):
This looks good, but is a bit hard to read. Please either  add some comments, or (better yet) use more informative names. For example
- C=validity mask, M = missing indices, O = observed idncies
- Mu_tilde = expeted sufficient statistics for mean, S_tilde = ESS for covariance

- I generally don't like multiple assignments per line. So I would replace the hard to read line
`M_i, O_i = M[i, ][M[i, ] != -1], O[i, ][O[i, ] != -1]`
with
`missing_ndx_i = (M[i,] != -1)`
`M_i = M[i, ][missing_ndx_i]`
etc
(I don't understand the meaning of `missing_ndx_i` TBH...)

- You don't need to keep a separate Mu_tilde and S_tilde for each i in a dict. This will take ON D*2) space. Instead you should update the running sum in place, so the ESS use O(D^2) space.

Drishttii:
Hello,

Thank you very much for the review. 

I apologize for not clarifying properly. Comments have been added, I hope they are sufficient and if not then please let me know.

M is the matrix that has -1 entries for missing entries and other values (up to dim of X_miss) for given entries. m_indx (M[i, ] != -1) is basically -1 where entries are missing in X_miss (X). 

I looked at the [gaussMissingFitEm.m](https://github.com/probml/pmtk3/blob/master/toolbox/BasicModels/gauss/sub/gaussMissingFitEm.m) and [emAlgo.m](https://github.com/probml/pmtk3/blob/master/toolbox/Algorithms/optimization/emAlgo.m). The later has kept max_iters = 100 and convergence parameter = 1e-4. I have updated both these similarly in my code. For complexity part of my code (according to matlab functions [gaussMissingFitEm.m](https://github.com/probml/pmtk3/blob/master/toolbox/BasicModels/gauss/sub/gaussMissingFitEm.m) and [emAlgo.m](https://github.com/probml/pmtk3/blob/master/toolbox/Algorithms/optimization/emAlgo.m)), it is O(rows*iterations_of_em_algo). Should this be improved further?  I could use some more suggestions on this. 

Thanks again :)

murphyk:
I don't think your code is right.
Read the math in sec 8.7.4.
In the E step, we accumulate the expected sufficient statistics as a running sum:
```
    EX(u) = mi;
    EX(o) = X(o, i);
    EXX(u, u) = EX(u) * EX(u)' + Vi;
    EXX(o, o) = EX(o) * EX(o)';
    EXX(o, u) = EX(o) * EX(u)';
    EXX(u, o) = EX(u) * EX(o)';
    ess.EXsum = ess.EXsum + EX;
    ess.EXXsum = ess.EXXsum + EXX;
```

In the M step, we just normalize the ESS:
```
function model = mstep(model, ess)
% Maximize
n = ess.n;
model.mu = ess.EXsum/n;
model.Sigma = ess.EXXsum/n - model.mu*model.mu';
end
```

So there is no need for Mtilde{i} and Stilde{i}, which will take O(ND) and O(ND^2) space. You can do it in constant space, independent of N, as shown in the matlab code. Also, it is incorrect to take the covariance of the imputed X's as your M step!

Drishttii:
I understood my mistake. There is no need to calculate `Mu_tilde` and` S_tilde`. Now I have taken the exact approach of the matlab code. However, I observed that while running the EM step for a number of iterations the running sum on mu (`EXsum`) and sigma (`EXXsum`) is going to increase on every iteration, ending up giving extremely large values. But I think this is not something we want. I could really use your help here. Thanks :)

Drishttii:
@murphyk please look into this one. Thanks again.

murphyk:
The sum of expected sufficient statistics will grow with the number of data points N (since it is a sum), but it should be reset to 0 at the beginning of each EM loop, so it should be independent of the number of iterations.
Your code seems to be learning the wrong results, judging by the plots. Unfortunately I don't have time to spot the bug for you. I suggest printing out all the numbers from the matlab code for a very small example (say N=3 data points, and T=2 iterations, in D=2 dimensions), and then checking you can reproduce them in python. (You will of course need to use the same dataset, so avoid random generation for this exercise.) Good luck!

Drishttii:
@murphyk Thank you for your reply. The bug was spotted and the sum of the expected sufficient stats now resets and it works well.

PR #356, opened 2021-04-19
Converted mixPpcaDemoNetlab to python

### LINKED ISSUE #230: Convert mixPpcaDemoNetlab to python
This is figure number 20.12 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/mixPpcaDemoNetlab.m)

shivaditya-meduri:
mixppca_k-1
mixppca_k-10

Closes #230,
I implemented the Probabilistic Principal Component Analysis mixture model following the research paper "Mixtures of Probabilistic Principal Component Analysers" by Michael E. Tipping and Christopher M. Bishop.
please visit pages 5-7 from  http://www.miketipping.com/papers/met-mppca.pdf for the mathematics and the algorithm used in this code

### DIFF
--- a/scripts/mixPpcaDemoNetlab.py
+++ b/scripts/mixPpcaDemoNetlab.py
@@ -2,21 +2,26 @@
 import math
 import matplotlib.pyplot as plt
 import numpy as np
-
-
+from scipy.special import logsumexp
 '''
 t = Wx + µ + E
-the equation above represents the latent variable model which relates a d-dimensional data vector t to a corresponding q-dimensional latent variables x with q < d, for isotropic noise E ∼ N (0, σ2I)
+the equation above represents the latent variable model which 
+relates a d-dimensional data vector t to a corresponding q-dimensional 
+latent variables x 
+with q < d, for isotropic noise E ∼ N (0, σ2I)
 t : latent
 x : data
 W : latent_to_observation matrix
 µ : centres_of_clusters
 E : var_of_latent
-Initially using Kmeans algorithm parameters are initialized before applying the expectation maximization algorithm for an estimate of the parameter values
+Initially using Kmeans algorithm parameters are initialized before 
+applying the expectation maximization 
+algorithm for an estimate of the parameter values
 '''
 
 
-def mixture_ppca_parameter_initialization(data, n_clusters, latent_space_dim, n_iterations):
+def mixture_ppca_parameter_initialization(data, n_clusters, latent_dim,
+                                          n_iterations):
     """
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
Hi @murphyk , @mjsML ,

I initially tried to implement directly from the gmm module of netlab, but I couldn't connect the Matlab code to the theory from the book. That's why I directly used the research paper http://www.miketipping.com/papers/met-mppca.pdf for reference.

Please review this PR.

Thank you

murphyk (changes requested):
This looks good, but is a bit hard to read. 
- It's good to use descriptive variable names for arguments to functions, but it makes the math hard to read, so I would shorten W=latent_to_observation, mu=centres_of_clusters, etc.
- Rather than writing out the full math expression, try to leverage known functions where possible, eg scipy.stats.multivariate_normal.log_pdf and scipy.special.logsumexp.
- Try to keep lines <= 80 characters.

shivaditya-meduri:
Hi @murphyk, Thank you for your valuable suggestions. I've incorporated them into the code.

murphyk (changes requested):
Looking much better, But a few nits:
- I don't think you need myMax etc if you use logsumexp - it takes cares of it (see sec 2.5.4 of my book)
- there are some typos (eg simga)
- the comment at the beginning is not quite correct - I think it describes the inference process for the E step, but not the generative model. See 20.2.6 of my book
- Please use z for latent variables, not t. (t is normally for target labels)

shivaditya-meduri:
Hi @murphyk, I've implemented the suggested changes. Thank you for your patience.

murphyk:
You still have a bunch of unnecessary MyMax commands. The point of logsumexp is that you don't need any of this!

Also your ellipses are much smaller than the original figure.
You should use [pyprobml_utils.confidence_ellipse](https://github.com/probml/pyprobml/blob/master/scripts/pyprobml_utils.py#L64) for plotting.

shivaditya-meduri:
Hi @murphyk , I tried using pyprobml_utils.confidence_ellipse but the figure came as follows:
image
this is because the method avoids the use of an iterative eigen decomposition algorithm which is used in the algorithm of the code. I observed that the confiden_ellipse is plotting the confidence interval of 2 standard deviation. So I changed from 1 sd to 2 sd in the code above and the ellipse seems to be larger. Please let me know if this doesn't work and I will start converting the matlab code directly from netlab to create an accurate figure as in the book. 
Thank you for your help.

PR #364, opened 2021-04-20
refactored matrix-factorization notebook

### LINKED ISSUE #357: refactor matrix-factorization notebook
The [matrix_factorization_recommender](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/matrix_factorization_recommender.ipynb) colab has 2 main parts: the first is pure numpy , the second uses the surprise library, since they are distinct.
The goal is to split this into 2 separate colabs.
Note:The first part  downloads extra data containing title and genre for each movie, so we can decode integer codes into something meaningful (see eg `get_movie_name` and `get_movie_genres`). This will need to be copied to the second colab.

Nirzu97:
@murphyk  , @mjsML  please review this.

I have refactored original notebook into two understandable notebooks.

closes #357

### DIFF
--- a/notebooks/matrix_factorization_recommender.ipynb
+++ b/notebooks/matrix_factorization_recommender.ipynb
@@ -2,17 +2,20 @@
   "nbformat": 4,
   "nbformat_minor": 0,
   "metadata": {
+    "colab": {
+      "name": "Untitled7.ipynb",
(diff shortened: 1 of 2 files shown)


murphyk:
LGTM, thanks

PR #372, opened 2021-04-21
Use jax.numpy consistently (iris_optimize.py)

Drishttii:
@murphyk and @mjsML please review this. Here, I have used both JAX and Torch.

#371


murphyk:
In colab, I get the error

`ValueError: Using a target size (torch.Size([2])) that is different to the input size (torch.Size([2, 1])) is deprecated. Please ensure they have the same size.`

Drishttii:
Yes. Also there was one more error. In the file `scipts/armijo_sgd.py` in the `ArmijoModel ``opt` is not required and used. So the step function should only return minibatch_loss or else you will get the `TypeError: unsupported operand type(s) for +=: 'float' and 'tuple' in line ‘loss_sum += loss’` in `iris_optimize.py` where this model is used. I have created another issue for change in this `scipts/armijo_sgd.py `file. Thank you for the comment.

Drishttii:
Yes, the step size is recorded. Is it required to be printed?

PR #381, opened 2021-04-21
Use jax.numpy consistently (rbf_2d_plot.py)

Drishttii:
@murphyk and @mjsML , please review this. Here, ‘import jax.numpy as jnp’ is commented out and not used. Is this required to be changed?

#380


murphyk:
This issue refers to `mean_shift_opt.py` but the code is `scripts/rbf_2d_plot.py`. 
Anyway, it's true that mean_shift comments out jax.numpy. The problem is that the code still uses
onp.foo() instead of np.foo(), where onp = original numpy. Please fix this.

Drishttii:
I am sorry I don't understand. The problem of onp.foo() instead of np.foo() is in which file? `scripts/mean_shift.py` or `scripts/rbf_2d_plot.py `

murphyk:
In rbf_2d_plot.py, I commented out jax, so this change is unnecessary.

Drishttii:
Alright. Got it. Thanks :)

PR #397, opened 2021-04-21
Use jax.numpy consistently (autodiff_demo.py)

Drishttii:
@murphyk and @mjsML please review this. Thanks :)

#396


murphyk (changes requested):
This code does not run in colab! I get an error with the line `tf.enable_eager_execution()`.

Drishttii:
Yes, thank you for pointing this out. This would work for both.

Issue #204, opened 2021-03-11
Convert ridgePathProstate to python


mjsML:
This is figure number 11.12 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/ridgePathProstate.m)

3devRapeti:
I want to work on this.

murphyk:
Just use [sklearn ridge regression](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html#sklearn.linear_model.Ridge) with warm-starting as you sweep over the range of lambda.

karalleyna:
I would like to work on this.

mjsML:
> I would like to work on this.

please go ahead.

PR #398, opened 2021-04-21
Create ridge regression path figures

karalleyna:
I have created two figures showing Ridge Regression path. 

Ekran Resmi 2021-04-21 23 53 05
Ekran Resmi 2021-04-21 23 53 11

I decided to use larger alpha values after reading the related part of the book in order for figures to be consistent those in the book. So, it can be changed according to your feedback. @mjsML

### DIFF
--- a/scripts/ridgePathProstate.py
+++ b/scripts/ridgePathProstate.py
@@ -1,0 +1,49 @@
+from sklearn.linear_model import Ridge
+from sklearn.linear_model import RidgeCV
+from scipy.io import loadmat
+from sklearn import linear_model
+import jax.numpy as jnp
+import numpy as np
+
+def get_weights(X, y, alphas):
+    w = np.zeros((len(alphas),X.shape[1]))
+    for i in range(len(alphas)):
+        w[i,:] = Ridge(alpha=alphas[i]).fit(X, y).coef_
+            return w
+
+
+data = loadmat('../data/prostate/prostateStnd')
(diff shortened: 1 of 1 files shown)


murphyk (changes requested):
Looks good, but please make a few changes
- just use numpy, no need to use jax in this case
- it's confusing to have alphas and alphas_cv, and you end up fitting the model more times than necessary. Please just use alphas_cv
- The dof on the x-axis for the first plot is different from the matlab code. I don't understand your equation: where does sum_d var(x_d)/(var(x_d) + alpha) come from?
- Best to just plot alpha on the x-axis rather than dof or integer index

murphyk:
Also, please link to the issue that you are solving, by saying 
closes #204

karalleyna:
The former figure in the code, the regression path was plotted like a figure whose x axis showed dof which I supposed it means the degrees of freedom of ridge regression, whose formula is also given in the book. Therefore, I plotted two figures, one is showing the weights against dof and the latter one was plotted via the transpose of the coefficient matrix since I tried to be consistent to the previous code as well as the figure in the book. Furthermore, since there are issues about the use of jax.numpy in colab, so I used jax.numpy even if the use of numpy is mentioned within the guidelines.

murphyk:
Where did you see the equation for dof in the book? 
Also, 
It seems that #410 also implements this ridge path code. Since their code is simpler, I will close this issue.

karalleyna:
I have found it at Chapter 11.3.2.

PR #409, opened 2021-04-22
Changes in step function of armijo_sgd.py

### LINKED ISSUE #408: Changes in step function in armijo_sgd.py
Step function in the ArmijoModel should simply return the minibatch_loss.

Drishttii:
@murphyk and @mjsML please check this change. Thank you.

Closes #408


murphyk:
are you sure ? i thought the armijo function was supposed to keep track of the step size it uses at each iteration, so you can track it?

Drishttii:
If the model is expected to track the step size then this change in the Armijo model would be unnecessary. Now in the `iris_optimize.py` I have made a change where step size is recorded. Is it required to be printed?

PR #411, opened 2021-04-22
Converted mixPpcaDemoNetlab to python

### LINKED ISSUE #230: Convert mixPpcaDemoNetlab to python

shivaditya-meduri:
mixppca_k-1
mixppca_k-10
Closes #230,
I implemented the Probabilistic Principal Component Analysis mixture model following the research paper "Mixtures of Probabilistic Principal Component Analysers" by Michael E. Tipping and Christopher M. Bishop.
please visit pages 5-10 from http://www.miketipping.com/papers/met-mppca.pdf for the mathematics and the algorithm used in this code
I made the suggested changes:
1. Used the log sum trick everywhere
2. Changed the confidence interval to 2 standard deviations like in confidence_ellipse

### DIFF
--- a/scripts/mixPpcaDemoNetlab.py
+++ b/scripts/mixPpcaDemoNetlab.py
@@ -1,0 +1,262 @@
+# Author: Meduri Venkata Shivaditya
+import math
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy.special import logsumexp
+'''
+z = Wx + µ + E
+the equation above represents the latent variable model which 
+relates a d-dimensional data vector z to a corresponding q-dimensional 
+latent variables x 
+with q < d, for isotropic noise E ∼ N (0, σ2I)
+z : latent
+x : data
+W : latent_to_observation matrix
+µ : centres_of_clusters
+E : var_of_latent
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
Hi @murphyk , I tried using pyprobml_utils.confidence_ellipse but the figure came as follows:
image
I think this is because the method avoids the use of an iterative eigen decomposition algorithm which is used in the algorithm of the code I implemented to plot the PC vectors and that's why they are not orienting correctly. I observed that the confidence_ellipse is plotting the confidence interval of 2 standard deviation. So I changed from 1 sd to 2 sd in the code above and the ellipse seems to be larger. Please let me know if this doesn't work and I will start converting the matlab code directly from netlab to create an accurate figure as in the book. 
Thank you for your help.

murphyk:
LGTM :). 2 sigma is more common since it covers 95%.

shivaditya-meduri:
Thank you @murphyk

PR #429, opened 2021-04-24
Use jax.numpy consistently (student_pgm_inf_autodiff.py)

Drishttii:
In `student_pgm_inf_autodiff.py`, `BayesNetInfAutoDiff` class of `bayesnet_inf_autodif.py` is used. I would like to know why in the `make_list_of_factors()` function the `evecs` (evidence vectors) and `cpts` (params) lists are merged? In line `return (evecs+cpts)`. Doing so gives the error “ValueError: Number of einsum subscripts must be equal to the number of operands” dag and params are passed in `bn.BayesNetInfAutoDiff(dag, params)`.  @murphyk please solve my query. Thanks.

### DIFF
--- a/scripts/student_pgm_inf_autodiff.py
+++ b/scripts/student_pgm_inf_autodiff.py
@@ -1,9 +1,7 @@
 """
 Inference in student network using autodiff compared to pgmpy.
-
 Network is from
 https://github.com/pgmpy/pgmpy_notebook/blob/master/notebooks/2.%20Bayesian%20Networks.ipynb
-
 I     D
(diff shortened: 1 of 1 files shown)


murphyk:
Good question. It seems this code does not really need jax, but it does not run even with vanilla numpy or pgmpy. I will close this issue, and open a new one, called "fix student_pgm_inf_autodiff", since it seems there are bigger problems than merely not using jax.numpy consistently. We can come back to this later.

Drishttii:
Thank you for notifying :)

PR #447, opened 2021-05-01
Convert mixexpDemoOneToMany to python

### LINKED ISSUE #213: Convert mixexpDemoOneToMany to python
This is figure number 13.25 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/mixexpDemoOneToMany.m)

Drishttii:
@murphyk and @mjsML please review this. I have tried to write all the Matlab equations from scratch. I will try to improve this further if needed. Thanks :)

Closes #213 

Figures:
a)
Inverse_problem
b)
Gating_functions
c)
Expert_predictions
d)
Predictions

### DIFF
--- a/scripts/mixexp.py
+++ b/scripts/mixexp.py
@@ -37,20 +37,14 @@ def is_pos_def(x):
 K = 3 #nmix
 D = np.size(X, axis=1)
 N = np.size(X, axis=0)
-np.random.seed(1)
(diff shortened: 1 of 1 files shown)


murphyk (changes requested):
There seems to be a problem since the final prediction is quite bad. Perhaps it's a local optimum for EM? The gating network seems to have trained, but not the experts themselves.

PR #448, opened 2021-05-03
Converted groupLassoDemo to python

### LINKED ISSUE #207: Convert groupLassoDemo to python
This is figure number 11.16 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/groupLassoDemo.m)

shivaditya-meduri:
Closes #207
W_groupLassoLinf_reg_piecewise_constant (1)
W_groupLassoL2_reg_piecewise_gaussian (1)
W_groupLassoL2_reg_piecewise_constant (1)
W_actual_piecewise_gaussian (1)
W_actual_piecewise_constant (1)
W_lasso_reg_piecewise_gaussian (1)
W_lasso_reg_piecewise_constant (1)
W_groupLassoLinf_reg_piecewise_gaussian (1)

### DIFF
--- a/scripts/groupLassoDemo.py
+++ b/scripts/groupLassoDemo.py
@@ -1,0 +1,124 @@
+# Author: Meduri Venkata Shivaditya
+"""
+Figure 11.16 and 11.17 in the book "Probabilistic Machine Learning: An Introduction by Kevin P. Murphy"
+Dependencies: spams(pip install spams), group-lasso(pip install group-lasso)
+
+Illustration of group lasso:
+
+To show the effectiveness of group lasso, in this code we demonstrate:
+a)Actual Data b)Vanilla Lasso c)Group lasso(L2 norm) d)Group Lasso(L infinity norm)
+on signal which is piecewise gaussian and on signal which is piecewise constant
+we apply the regression methods to the linear model - y = XW + ε and estimate and plot W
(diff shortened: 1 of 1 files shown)


shivaditya-meduri:
Hi @murphyk, Please review this PR. Thank you for your time.

shivaditya-meduri:
Please note that the y - axis ranges are different for different graphs

murphyk (changes requested):
This looks good, but I think you forgot the 'debiasing' step at the end. Basically this uses the sparse estimate to identify which coefficients are 0 and which are not-zero, and then computes the MLE for the non-zero coefficients, so they don't get "shrunken". This should ensure the resulting (non-zero) coefficients are all on the same vertical scale regardless of the method used to estimate the support of the vector.

shivaditya-meduri:
Hi Sir, I initially tried to work with in-built libraries, but noticed that they all included relaxed lasso implementation by including a 0<phi<1 multiplied with the regularization param or just included an additional L1 term, I couldn't find an implementation of 2 stage estimation method of debiasing online, so I implemented it basing on the book. I am not sure whether I've implemented it correctly, Please do let me know if I haven't implemented it correctly and I will implement it the right way.
Thank you for your confidence.

murphyk:
this looks great, thanks!

Issue #437, opened 2021-04-26
convert pymc3 demo to pyro for beta-binomial model


murphyk:
Rewrite [this script](https://github.com/probml/pyprobml/blob/master/scripts/prior_post_pred_binom_pymc3.py) to use pyro (or numpyro) instead of pycm3.

PR #449, opened 2021-05-03
Removed pymc3 and used pyro

VaibhaviMishra04:
resolves the issue https://github.com/probml/pyprobml/issues/437.

### DIFF
--- a/scripts/prior_post_pred_binom_pymc3.py
+++ b/scripts/prior_post_pred_binom_pymc3.py
@@ -1,37 +1,59 @@
-# prior and posterior predctiive for beta binomial
-# fig 1.6 of 'Bayeysian Modeling and Computation'
-
 import arviz as az
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
-import pymc3 as pm
+import torch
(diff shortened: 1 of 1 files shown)


murphyk (changes requested):
I get the following error when I run this in colab (at the mcmc.run(Y) command)
```
ValueError: Input arguments must all be instances of numbers.Number, torch.Tensor or objects implementing __torch_function__.
Trace Shapes:
 Param Sites:
Sample Sites:
```

Please use numpyro instead of pyro (sorry, I was not clear).

Please make a new file prior_post_pred_binom_numpyro.py, rather than editing the old pymc3 file.

murphyk:
No need to solve this, it has been done.

VaibhaviMishra04:
Sorry, I tried but I couldn't get the graph for posterior predictive distribution and prior predictive distribution wrong, I was able to plot the other two correctly[using numpyro]. I am unable to find the correct solution to this issue. Would it be possible for you to drop the link to the right solution?

murphyk:
No worries, I think we can stick to the pymc3 version for now.

PR #453, opened 2021-05-10
Create Sparsennetdemo.py

### LINKED ISSUE #210: Convert sparseNnetDemo to python
This is figure number 13.21 in the book. its matlab code is [here](https://github.com/probml/pmtk3/blob/master/demos/sparseNnetDemo.m)

karalleyna:
Closes #210

I have implemented iterative soft-thresholding algorithm in order to obtain Figure 13.21(b) in the book.

indir

I have just changed the data synthesis part of the original code. Could you please review it? @murphyk @mjsML


karalleyna:
I changed the data synthesis considering the original code and also added the code related to the diagram of sparse neural network. Figures I obtained are 
Neural Network gv
indir

murphyk:
I cannot reproduce your example - it seems there is some non-determinism somwhere. Maybe just use jax RNG exclusively, and not mix with np.random.

karalleyna:
> I cannot reproduce your example - it seems there is some non-determinism somwhere. Maybe just use jax RNG exclusively, and not mix with np.random.

I removed all np.random in generate_data.

PR #464, opened 2021-05-20
Visualize posterior marginals of a discrete graphical model

### LINKED ISSUE #458: visualize posterior marginals of a discrete graphical model
The [asia_pgm.ipynb](https://github.com/probml/pyprobml/blob/master/notebooks/asia_pgm.ipynb) notebook computes posterior marginals in a small, discrete graphical model (known as the "Asia" network) using the [pgmpy package](https://pgmpy.org/). The task is to visualize these posterior marginals on top of the graph itself, similar to the picture below. 
(This corresponds to 4 different evidence sets, as explained in the colab.)

Screen Shot 2021-05-19 at 12 24 12 PM

Since rendering histograms may be hard, a simpler solution is just to print the marginals in text form inside the nodes.
eg replace the string "Asia" with "Asia (0.5, 0.5)", to mean that there is 50% probability that Asia=0, and 50% probability that Asia=1. The main tricky bit is to figure out how to modify the dot file generated by [graphviz](https://pypi.org/project/graphviz/). The [numpyro site](https://numpyro.readthedocs.io/en/latest/tutorials/model_rendering.html) has some relevant material.

The code should work for any graph structure. In particular, after getting it working on the Asia network, please also apply it to the [student network](https://github.com/probml/pyprobml/blob/master/notebooks/student_pgmpy.ipynb).

Drishttii:
@murphyk and @mjsML please review this. Numpyro allows visualizing by creating plates, but I found the graphviz feature of representing each node as an `HTML` table more appealing. I believe this will help in looking at the student_network model better, which has tables to be visualized at each node. 

Thanks :)

Closes #458

### DIFF
--- a/notebooks/asia_pgm.ipynb
+++ b/notebooks/asia_pgm.ipynb
@@ -1,0 +1,865 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "Copy of asia_pgm.ipynb",
+      "provenance": [],
+      "toc_visible": true
+    },
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3"
+    },
+    "language_info": {
+      "name": "python"
+    }
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "caeOoFYBb5p2"
+      },
+      "source": [
+        "# The \"Asia\"  graphical model\n",
+        "\n",
+        "We illustrate inference in the \"Asia\" medical diagnosis network.\n",
+        "Network is from http://www.bnlearn.com/bnrepository/#asia.\n"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "colab": {
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "p4SVMjkpbcJ7",
+        "outputId": "1eaf5218-69cf-44b1-e7a9-4c19f872317e"
+      },
+      "source": [
+        "!pip install  causalgraphicalmodels\n",
+        "!pip install pgmpy\n",
+        "\n",
+        "#!ls /usr/local/lib/python3.7/dist-packages/pgmpy/utils"
+      ],
+      "execution_count": null,
+      "outputs": [
+        {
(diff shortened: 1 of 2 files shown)


murphyk:
This looks good. But please just use 2 decimal places. Also please rename the file to asia_pgm2.ipynb so it doesn't conflict with my current file. (Doing version control with notebooks is tricky)

Drishttii:
Shall I delete the asia_pgm.ipynb?

murphyk:
I've taken your visualization code and factored it out from the notebook and added it to a new [pgmpy_utils.py file](https://github.com/probml/pyprobml/blob/master/scripts/pgmpy_utils.py). I  modified my marginals code it so it returns the full distribution for each node, not just the probability of state 1, and added it there too. I rewrote [asia_pgm.ipynb](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/asia_pgm.ipynb) to use these new functions.  These changes actually revealed a subtle bug in your code: you were assuming 'yes' was state 1, but in fact it was state 0. I have changed this in my code.

However, there is more to do - currently your visualize_marginals code assumes there are only 2 states, and assumes they are called 'yes' and 'no', but if you look at [student_pgmpy.ipynb](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/student_pgmpy.ipynb), you will 
see an example of a model where some nodes have  3 states, and each state has a different name. Please can you generalize your plotting code so the html table shows K lines, where K is the number of states, and it uses the appropriate names.

Please add your code changes to [pgmpy_utils.py file](https://github.com/probml/pyprobml/blob/master/scripts/pgmpy_utils.py). This way, the existing [asia_pgm.ipynb](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/asia_pgm.ipynb) should automatically inherit your fixes. Then make a new version of [student_pgmpy.ipynb](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/student_pgmpy.ipynb) that calls your visualization code, and make sure it works (ie both asia and student notebooks should both call pgmpy_utils.visualize_marginals)

Drishttii:
Added the minor changes for calling `pgmpy_utils.py` functions.

murphyk:
This will be solved by #473

PR #473, opened 2021-05-24
Changes in pgmpy_utils.py file

Drishttii:
@murphyk and @mjsML please have a look at this. I have added the required functions here.

### DIFF
--- a/scripts/pgmpy_utils.py
+++ b/scripts/pgmpy_utils.py
@@ -12,6 +12,13 @@ def get_state_names(model, name):
   state_names = cpd.state_names        
   return state_names
 
+def get_all_state_names(model):
+  state_names = dict()
+  for cpd in model.get_cpds():
+    for k, v in cpd.state_names.items():
+      state_names[k] = v
+  return state_names
+
 def get_lengths(state_names, name):
   row = []
   for k, v in state_names.items():
@@ -19,11 +26,6 @@ def get_lengths(state_names, name):
       col = len(v)
     else:
       row.append(len(v))
-  return row, col
-
-def get_length(state_names, name):
-  row = len(state_names[name])
-  col = row
   return row, col
 
 def get_values(model, name):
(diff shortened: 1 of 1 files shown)


murphyk:
- Your `get_state_names(model, name)' has a different API than my `get_state_names(model)` and breaks the `get_marginals` function...
 - Also, for very trivial functions, like ` get_length(state_names, name)`, I suggest you define them as subfunctions of `visualize_model`, or just inline the code

murphyk:
I used your `visualize_model` function in the [student_pgm example ](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/student_pgmpy.ipynb#scrollTo=LXTnimbBf3yy) and I noticed the rows of the L node at the end have funny labels. 

Please use 2 decimal places for all numbers.

Also, rewrite so the rows are labeled by the state names of their parents, as in the example G node on the right in the figure below.

Screen Shot 2021-05-24 at 11 37 43 AM

Drishttii:
Hopefully, I have addressed all the above things, but do let me know if there is anything missing :)

murphyk:
Almost perfect :). A few minor issues.

- The `visualize_marginals` function was missing the `evidence` argument. Please change API to `visualize_marginals(model, evidence, marginals)`
 
- When you display marginals for the observed G node in the example below, it has 2 identical columns. But the martginal distribution over values should have 1 entry for each state (with 1.0 on the obsrerved value, and 0.0 on the rest). Example (from [this colab](https://colab.research.google.com/github/probml/pyprobml/blob/master/notebooks/student_pgmpy.ipynb#scrollTo=GvQFdruIvlu8)) 
Screen Shot 2021-05-27 at 10 16 30 AM

- When you display marginals, please use a single row with K columns, rather than K rows with 1 column.
(This will match the appearance of display_model root node CPTs.)

-For visualize model, when you display the combinations of values for each parent state on the left column, please remove the (brackets) and please insert comma between entries. eg in example below, write "dumb, easy", etc
Screen Shot 2021-05-27 at 10 19 47 AM

- There is no need to create funcitons for 'one-liners' like
```
def get_row_string(input):
  string = "<TR>" + str(input) + "</TR>"
  return string
```
So you can replace
```
 row_string = row_string + get_row_string(col_string)
```
with
```
 row_string = row_string + ("<TR>" + str(col_str) + "</TR>")
```
Similarly for `get_column_string` and `get_values`.
This reduces the number of trivial functions.

Drishttii:
Done! And changes in `student_pgmpy.ipynb` and` asia_pgm.ipynb` while calling` visualize_marginals` function taking the argument `evidence` (calling `visualize_marginals(model, evidence, marginals)`) have to be made.

murphyk:
thanks.  i will update the 2 notebooks.
(Also, I will change `visualize_marginal(marginals, evidence, model)` to `visualize_marginals(model, evidence, marginals)` to be consistent with `get_marginals(model, evidence)` :)

PR #474, opened 2021-05-24
Visualize CPTs for discrete graphical model

### LINKED ISSUE #459: visualize CPTs for discrete graphical model
In the [student_pgmpy.ipynb](https://github.com/probml/pyprobml/blob/master/notebooks/student_pgmpy.ipynb), we show the conditional probability tables (CPTs) for each node in graphical model, in addition to the graph structure, as shown below.

student-dgm

The goal of this task is to automate this process. The graph rendering can be done using graphviz, as illustrated in the notebook. The challenge is to figure out how to attach extra text information to the resulting graph (perhaps as dummy nodes?).

This issue is related to #458 but different - that issue just shows the nodes and their posterior marginals.
This issue is concerned with showing the parameters of the network.

Drishttii:
@murphyk and @mjsML please have a look at this.

I tried a lot to generalize the visualization for both the models, but I think in `asia` model we are trying to look at the marginals whereas we want to visualize the model of `student`. So I added two functions one for looking at the model and other for after calculating the marginals. I used the same `visualize_model` function to look at both `asia` and `student` model. 

Closes #459

### DIFF
--- a/notebooks/student_pgmpy.ipynb
+++ b/notebooks/student_pgmpy.ipynb
@@ -1,0 +1,362 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "Copy of Copy of student_pgmpy.ipynb",
+      "provenance": [],
+      "toc_visible": true
+    },
+    "kernelspec": {
+      "name": "python3",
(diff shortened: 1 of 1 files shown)


murphyk:
This will be solved by #473

PR #591, opened 2021-07-30
Create `LinearGaussianStateSpaceModel` class

gerdm:
## Description

* Create `state_space_models_lib.py` that implements the class `LinearGaussianStateSpaceModel` following Tensorflow's [LinearGaussianSpaceModel](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/LinearGaussianStateSpaceModel) class using distrax.
* Create `linear_gauss_model_demo.py` that showcases the use of the `LinearGaussianStateSpaceModel`

I propose creating a new file (as opposed to merging it with `nlds_lib.py`) since `nlds_lib.py` works "out of the box" with Jax installed, and `state_space_models_lib.py` required distrax. 

### Figures

image

### Issue 

You can link to the issue you have solved.

Closes [issue 311](https://github.com/probml/hermes/issues/311) in Hermes


scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+    if isinstance(seed, IntLike.__args__):
+        rng = jax.random.PRNGKey(seed)
+    else:  # key is of type PRNGKey
+        rng = seed
+
+    return rng, sample_shape
+
+
+class LinearGaussianStateSpaceModel(jittable.Jittable):
+    def __init__(self, num_timesteps, transition_matrix, transition_noise,
```
murphyk, line 51: I don't like the design in which num_timesteps is provided to the class constructor.
I know we need it to define the event shape, but we should also be able to handle sequences of arbitrary length.
I suggest you rename to to max_num_timesteps, and allow it to be None. (Some methods below may need changing)

scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+        self._observation_matrix = observation_matrix
+        self._observation_noise = observation_noise
+        self._initial_state_prior = initial_state_prior
+        self.observation_size, self.state_size = observation_matrix.shape
+
+    @property
+    def num_timesteps(self) -> Array:
+        return self._num_timesteps
+
+    @property
```
murphyk, line 65: why do you need all these property methods? why not access the fields directly?

scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+        self._num_timesteps = num_timesteps
+        self._transition_matrix = transition_matrix
+        self._transition_noise = transition_noise
+        self._observation_matrix = observation_matrix
+        self._observation_noise = observation_noise
+        self._initial_state_prior = initial_state_prior
+        self.observation_size, self.state_size = observation_matrix.shape
+
+    @property
+    def num_timesteps(self) -> Array:
```
murphyk, line 62: max_num_timesteps

scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+        obs_noise = self.observation_noise.sample(seed=key_obs) 
+
+        state_next = state_mean_next + state_noise
+        obs_next = self.observation_matrix @ state_next + obs_noise
+
+        return state_next, (state_next, obs_next)
+    
+    def _sample_n(self, key: PRNGKey, n: int) -> Tuple[Array, Array]:
+        key_init, key_next, key_step = random.split(key, 3)
+        key_steps = random.split(key_step, n * self.num_timesteps).reshape(n, self.num_timesteps, -1)
```
murphyk, line 154: you should pass num_timesteps as an  argument

scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+        state_init = self.initial_state_prior.sample(seed=key_init, sample_shape=(n,))
+        
+        vmap_sample = jax.vmap(lambda states, keys: jax.lax.scan(self.__sample_step, states, keys))
+        _, (state_samples, obs_samples) = vmap_sample(state_init, key_steps)
+
+        return state_samples, obs_samples
+
+    def sample(self, *,
+               seed: Union[IntLike, PRNGKey],
+               sample_shape: Union[IntLike, Sequence[IntLike]] = ()) -> Array:
```
murphyk, line 165: you should pass num_timesteps as an optional argument, which defaults to self.max_num_timesteps if not specified. Then the event shape is (H*T, D*T) for hidden dim H, obs dim D, and num timesteps T

scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+            Means of the smoothed marginal distributions
+        * array(*batch_size, timesteps, state_size, state_size)
+            Covariances of the smoothed marginal distributions
+        """
+        filtered_means, filtered_covs, mu_cond_hist, Sigma_cond_hist = self.forward_filter(x)
+        smoothed_means, smoothed_covs = self.backward_smoothing_pass(filtered_means, filtered_covs, mu_cond_hist, Sigma_cond_hist)
+        return smoothed_means, smoothed_covs
+
+    def log_prob(self, x: Array) -> Array:
+        """
```
murphyk, line 292: This code is very hard to read. You should compute Z(t) at each step from the normalization constant of the KF step. Then you can just sum  log Z across time. Logprob method can then just use output of the filtering / forwards algorithm.

(In the equation below I am missing an = sign)

Screen Shot 2021-07-30 at 3 06 22 PM

murphyk (changes requested):

murphyk:
FYI, [here](https://github.com/probml/pmtk3/blob/master/toolbox/LatentVariableModels/lds/sub/kalmanFilter.m#L96) is how I compute the logprob in my matlab kf code from 3 decades ago :)

Also,  I thought of a cool unit test for testing kalman smoothing that would also have pedagogical value.
Consider a small LG-SSM of length T=2, with hidden states of size H=3 and observed states of size O=4.
We can explicitly compute the joint distribtuion $p(z1, z2,  x1, x2) = p(z1) p(z2|z1) p(x1|z1) p(x2|z2)$ using the rules for combining linear Gaussians:

Screen Shot 2021-07-31 at 7 06 51 AM

Now you can condition this joint on observing x1 and x2, and compute the posterior marginals p(z1|x1,x2) and p(z2|x1,x2) using the standard rules for conditoning  Gaussians: 

Screen Shot 2021-07-31 at 7 07 51 AM

These marginals should match the results of Kalman smoothing.
(You could also test kalman filtering by checking that you get p(z1|x1) but p(z2|x1,x2).)

scripts/state_space_models_lib.py
```diff
@@ -0,0 +1,316 @@
...
+        seed: PRNG key or integer seed.
+        sample_shape: Additional leading dimensions for sample.
+
+        Returns
+        -------
+        * Array(*sample_shape, batch_shape, event_shape)
+            A sample of shape `sample_shape` + `batch_shape` + `event_shape`.
+        """
+        rng, sample_shape = convert_seed_and_sample_shape(seed, sample_shape)
+        num_samples = functools.reduce(operator.mul, sample_shape, 1)  # product
```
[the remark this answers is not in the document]
gerdm: I took this function directly from the class `Distribution` in [distribution.py](https://github.com/deepmind/distrax/blob/6da616249953dbf6aaefaeed4a8c4394e64887e4/distrax/_src/distributions/distribution.py#L115) in the Distrax library. I was originally inheriting from `distrax.Distribution`, which required me to implement the `_sample_n` method, and the `sample` method called `sample_n`. Since we know inherit from `jittable.Jittable` I took it from their class

scripts/state_space_models_lib.py
gerdm: I was following the TFP API. I also don't think it is necessary. Maybe we should  just not use it altogether?

scripts/state_space_models_lib.py
gerdm: To have immutable (ish) properties. It is a good convention  (see [property function](https://docs.python.org/3/library/functions.html#property))

This raises an error if the user writes 

```python
model = LinearGaussianStateSpaceModel(...)
model.transition_matrix = jnp.zeros((2, 2)) # This fails
```

PR #593, opened 2021-08-03
fix `nlds_lib.py`

gerdm:
## Description

* Fix bug for the `ExtendedKalmanFilter` and `ContinuousExtendedKalmanFilter` classes: incorrect computation of the filtering step
* Add resampling of points for the bootstrap filtering (as explained in Algorithm 7.5 of  Särkkä)

### Figures

The figures change as follows (left is old, right is new):

Output of `ekf_vs_ukf_demo.py`
<p float="left">
  elf-old
  ekf-new
</p>

Output of `bootstrap_filter_demo.py`

<p float="left">
  bootstrap-old
  bootstrap-new
</p>

### DIFF
--- a/scripts/bootstrap_filter_demo.py
+++ b/scripts/bootstrap_filter_demo.py
@@ -51,9 +51,10 @@ if __name__ == "__main__":
     model = ds.NLDS(lambda x: fz(x, dt), fx, Qt, Rt)
     sample_state, sample_obs = model.sample(key, x0, nsteps)
 
+    n_particles = 3_000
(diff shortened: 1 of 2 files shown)


murphyk:
The bootstrap_filter_demo seems to have a bug

Screen Shot 2021-08-05 at 5 19 13 AM

gerdm:
I just ran the code in Colab. It looks good on my side. (using`jax==0.2.17 `)

image

I think the bug occurs if the number of particles is low. I increased the  number of particles in the `bootstrap_filter_demo `

PR #594, opened 2021-08-04
mix PPCA mfa helper functions

Drishttii:
## Description

Helper functions for mix PPCA.


murphyk:
remember to always cite your sources!!

PR #597, opened 2021-08-10
Create `adf_logistic_regression_demo.py`

gerdm:
## Description

* Create demo comparing online training of a logistic regression model using ADF vs MCMC / Laplace (offline) batch training

### Figures

<p float="left">

</p>

### Issue 

Closes [Issue 279](https://github.com/probml/hermes/issues/279) on Hermes

### DIFF
--- a/scripts/adf_logistic_regression_demo.py
+++ b/scripts/adf_logistic_regression_demo.py
@@ -1,5 +1,7 @@
-# Example of the training of a logistic regression
-# model using Assumed Density Filter (ADF)
+# Online training of a logistic regression model
+# using Assumed Density Filtering (ADF).
+# We compare the ADF result with MCMC sampling
+# of the posterior distribution
 # Dependencies:
 #   !pip install git+https://github.com/blackjax-devs/blackjax.git
 #   !pip install jax_cosmo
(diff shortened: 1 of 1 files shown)


murphyk:
- modify data so that red and blue lines (w1 and w2) have different values, eg make decision boundary slope differently
- run for 50 observations to see if convergence is better with more data
- initial prior seems too narrow, the error bars should get smaller with more data (and converge to the batch /offlne estimate)
- code is hard to read, eg in mt_func, `log_term = log_term - (eta -  mu) ** 2 / (2 * v ** 2)` should say something like `log_term = log_term + jax.gauss.logpdf(mu, v)` or whatever.
- please use jnp.log1p(x) instead of jax.log(1+x)
- add citation to [Zoeter ADF  paper](https://ieeexplore.ieee.org/document/4383733) and annotate the code with eqn numbers
-  use more descriptive variable names, eg E=energy, E_base=energy_factory, M=ndims, N=ndata, q=prior_variance, etc
-  Add some comments, eg what are the lower/upper bounds, Phi=design matrix, "mij,sm->sij" maps m-dimensional feautres on (i,j) grid times (s,m) samples to get (s,i,j) grid of predictions.

PR #598, opened 2021-08-11
hclust_yeast_demo

Drishttii:
## Description

Data from probml-data.

## Checklist:

- [x] Tested on Google Colab.
- [x] Made a public PR


murphyk:
you should not use `!wget foo` in python scripts, it only works in colab, but will not run anywhere else.
Follow the example at https://github.com/probml/pyprobml/blob/master/scripts/ridgePathProstate.py

PR #601, opened 2021-08-11
tmp

Drishttii:
## Description

Data from probml-data

## Checklist:

- [x] Tested on Google Colab.
- [x] Made a public PR


murphyk:
tmp.py is a temporary file that you don't need to convert

PR #641, opened 2021-09-01
create: `thompson_sampling_bernoulli.py`

gerdm:
## Description

Create demo of multi-armed bandit problem using Thompson sampling. Each figure is stored in an individual filename `thompson_sampling_bernoulli_w{k}_t{t}.pdf`, where `k` is the index of the arm and `t` is the iteration number. 

### Figures
Putting all the figures together results in the following image

image

### DIFF
--- a/scripts/thompson_sampling_bernoulli.py
+++ b/scripts/thompson_sampling_bernoulli.py
@@ -10,16 +10,30 @@ from jax import random
 from jax.nn import one_hot
 from jax.scipy.stats import beta
 
+
+class BetaBernoulliBandit:
+    def __init__(self, alpha0, beta0, num_bandits):
+        self.alpha0 = alpha0
+        self.beta0 = beta0
+        self.num_bandits = num_bandits
+    
+    def sample(self, key, alphas, betas):
+        actions = random.beta(key, alphas, betas)
+        return actions
+    
+    def predict_reward(self, actions):
+        rewards = actions.argmax()
+
 def thompson_sampling_step(state, key):
     alphas, betas, reward_per_arm = state
-    keys = key
     K = len(alphas)
+    key_action, key_reward = random.split(key)
     
     # Choose an arm to pull
     # (Sample from the policy distribution)
-    action_t = random.beta(key, alphas, betas).argmax()
+    action_t = random.beta(key_action, alphas, betas).argmax()
     # Pull the arm and observe reward (either 1 or 0)
-    reward = random.bernoulli(key, reward_per_arm[action_t])
+    reward = random.bernoulli(key_reward, reward_per_arm[action_t])
     
     # Update policy distribution
     ind_vector = one_hot(action_t, K)
--- a/scripts/thompson_sampling_bernoulli.py
+++ b/scripts/thompson_sampling_bernoulli.py
@@ -11,6 +11,7 @@ from jax.nn import one_hot
 from jax.scipy.stats import beta
 from functools import partial
 
+
 class BetaBernoulliBandits:
     def __init__(self, K):
         self.K = K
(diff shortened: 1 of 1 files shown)


scripts/thompson_sampling_bernoulli.py
```diff
@@ -0,0 +1,66 @@
...
+from jax.scipy.stats import beta
+
+def thompson_sampling_step(state, key):
+    alphas, betas, probs = state
+    keys = key
+    K = len(alphas)
+    
+    # Choose an arm to pull
+    # (Sample from the policy distribution)
+    theta_t = random.beta(key, alphas, betas).argmax()
```
murphyk, line 20: change theta_t to action_t

scripts/thompson_sampling_bernoulli.py
```diff
@@ -0,0 +1,66 @@
...
+import jax
+import jax.numpy as jnp
+import matplotlib.pyplot as plt
+import pyprobml_utils as pml
+from jax import random
+from jax.nn import one_hot
+from jax.scipy.stats import beta
+
+def thompson_sampling_step(state, key):
+    alphas, betas, probs = state
```
murphyk, line 14: Maybe change `probs` to `mean_rewards` or `true_rewards`?

scripts/thompson_sampling_bernoulli.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+p_range = jnp.linspace(0, 1, 100)
+colors = ["orange", "blue", "green", "red"]
+colors = [f"tab:{color}" for color in colors]
+
+T = 200
+key = random.PRNGKey(314)
+keys = random.split(key, T)
+probs = jnp.array([0.65, 0.4, 0.5, 0.9])
```
murphyk, line 39: change `probs` to `reward_per_arm` or something like that

scripts/thompson_sampling_bernoulli.py
```diff
@@ -0,0 +1,66 @@
...
+
+import jax
+import jax.numpy as jnp
+import matplotlib.pyplot as plt
+import pyprobml_utils as pml
+from jax import random
+from jax.nn import one_hot
+from jax.scipy.stats import beta
+
+def thompson_sampling_step(state, key):
```
murphyk, line 13: I suggest a slightly more general API (as a warmup to the contextual case :).
Something like this.
```
def thompson_sampling_step(key, bel, true_reward, predict_reward):
  params = bel.sample(key)
  action = predict_reward(params).argmax()
  reward = true_reward(action)
  bel = bel.update(action, reward)
  return bel
```

You can then implement bernoulli bandit and Gaussian bandits with the same API, but diffferent update functions. Contextual bandits will be a small extension (future PR).

murphyk (changes requested):

murphyk:
Also, maybe use distrax rather than hard-coding all the conjugate equations?

murphyk:
Maybe a better API is this:
```
def thompson_sampling_step(key, bel, true_reward):
  params = bel.sample(key)
  pred_rewards = [bel.predict_reward(params, a) for a in bel.actions]
  action = pred_rewards.argmax()
  reward = true_reward(action)
  bel = bel.update(action, reward)
  return bel
```
Then we can define
```
class BetaBernoulliBandit:
  def init(alpha0, beta0):
     ...
  def sample(self, key):
     return  random.beta(key, self.alphas, self.betas)

   def update(self, action, reward):
     if reward == 0:
       self.alphas[action] += 1
     else:
     self.betas[action] += 1
 
  def predict_reward_bayes(self, a): // posterior mean
      return self.alphas(a)/(self.alphas(a) + self.betas(a))

  def predict_reward(self,  params, a): 
      return params[a]  //could be a neural network!
```

PR #700, opened 2022-03-22
Resolves #692

### LINKED ISSUE #692: exact 1d Bayesian filtering
Replicate the examples in the paper below using JAX.

@INPROCEEDINGS{Roth2017icassp,
  title     = "Computation and visualization of posterior densities in scalar
               nonlinear and non-Gaussian Bayesian filtering and smoothing
               problems",
  booktitle = icassp,
  url = 
https://liu.diva-portal.org/smash/get/diva2:1173608/FULLTEXT02.pdf
  author    = "Roth, M and Gustafsson, F",
  pages     = "4686--4690",
  month     =  mar,
  year      =  2017
}

benlau6:
### Figure Number

[bayesian filtering paper](https://liu.diva-portal.org/smash/get/diva2:1173608/FULLTEXT02.pdf) - 1, 2, 3, 4

### Figures

[gist](https://gist.github.com/benlau6/52e398bc4cfbdc36306fa2c4295b7e1a)

### Issue

[#692](https://github.com/probml/pyprobml/issues/692)

## Steps

- [x] Implemented point pass filters and smoothers
- [x] Implemented novel algorithm for Bayesian filtering
- [x] Created 1 \*utils.py file for generalized functions, and 3 \*example\*.py files for each of the examples

## Checklist

- [x] Tested on Google Colab.

Data from original paper are random samples which cannot be identified. "plot_densities" helper function in \*utils.py is created for screening weird plots as in paper. As a result, a new set of data and plots are generated as shown in gist to resemble.


benlau6:
@murphyk, @mjsML and @gerdm My code is ready, please review the code.

scripts/exact_1d_bayesian_filtering_utils.py
```diff
@@ -0,0 +1,384 @@
...
+    x = x_grid[idx]
+    return x + noise
+
+
+def kde(x_grid, x, kernel_variance):
+    delta = x_grid[1] - x_grid[0]
+
+    px = jax.vmap(functools.partial(jsp.stats.norm.pdf, loc=x, scale=kernel_variance))(
+        x_grid
+    )
```
gerdm, line 66: Better if you divide this in two/three steps (lines of code): the first one defines the partial function, then you define the vmap, and then you apply the vmap to get `px`. 
Also, I'm not sure if you need to vmap this function and apply a partial to it. You should be able to do that directly from jax.scipy.stats.

scripts/exact_1d_bayesian_filtering_example2.py
```diff
@@ -0,0 +1,76 @@
...
+    grid_maxval=grid_maxval,
+    num_grid_points=num_grid_points,
+    x0_rvs=x0_rvs,
+    v_rvs=v_rvs,
+    e_rvs=e_rvs,
+    f=f,
+    h=h,
+    K=K,
+    plot_xy=False,
+)
```
gerdm, line 58: Try not to break the lines of code too much.

scripts/exact_1d_bayesian_filtering_example3.py
```diff
@@ -0,0 +1,79 @@
...
+
+# to get x from measurement without noise
+def inv_h(y):
+    return y
+
+
+# functions to get sample
+v_rvs = lambda rng_key, shape: random.normal(rng_key, shape=shape)
+e_rvs = lambda rng_key, shape: random.normal(rng_key, shape=shape) * jnp.sqrt(0.5)
+x0_rvs = lambda rng_key, shape: random.normal(rng_key, shape=shape) * jnp.sqrt(0.1)
```
gerdm, line 33: It's better to use `def` even if defining one-liner functions. See: https://peps.python.org/pep-0008/#programming-recommendations

scripts/exact_1d_bayesian_filtering_utils.py
```diff
@@ -0,0 +1,384 @@
...
+    y,
+    x_grid,
+    x0_pdf,
+    v_rvs: Callable,
+    e_rvs: Callable,
+    f: Callable,
+    h: Callable,
+    num_samples: int,
+    K,
+    kernel_variance,
```
gerdm, line 83: Don't break the lines of the function after defining every parameter of the function.

scripts/exact_1d_bayesian_filtering_example3.py
```diff
@@ -0,0 +1,79 @@
...
+import exact_1d_bayesian_filtering_utils as utils
+
+
+# state transition function
+def f(x, v, k=None):
+    return 0.7 * x + v
+
+
+# measurement function
+def sat(x, minval, maxval):
```
gerdm, line 17: Rename function to something more meaningful.

scripts/exact_1d_bayesian_filtering_example3.py
```diff
@@ -0,0 +1,79 @@
...
+from jax import numpy as jnp
+from jax import scipy as jsp
+from jax import random
+import matplotlib.pyplot as plt
+
+import exact_1d_bayesian_filtering_utils as utils
+
+
+# state transition function
+def f(x, v, k=None):
```
gerdm, line 12: Rename function to something more meaningful.

scripts/exact_1d_bayesian_filtering_example3.py
```diff
@@ -0,0 +1,79 @@
...
+x0_pdf = functools.partial(jsp.stats.norm.pdf, scale=jnp.sqrt(0.1))
+
+# input arguments
+rng_key = random.PRNGKey(0)
+num_samples = 10000
+grid_minval = -6
+grid_maxval = 6
+num_grid_points = 500
+K = 24
+k = 18
```
gerdm, line 46: What is `K`, and what `k`? Rename.

scripts/exact_1d_bayesian_filtering_utils.py
```diff
@@ -0,0 +1,384 @@
...
+    x_pdf,
+    v_pdf,
+    e_pdf,
+    f,
+    h,
+    inv_h,
+    K,
+    k,
+    plot_all_densities=False,
+    title="",
```
gerdm, line 307: Don't break lines too much.

gerdm (changes requested):
Some comments:
* Try to use more meaningful names for your variables
* Don't break the lines too much when defining and using functions
* Replace `lambda` functions for `def`s (see Pep-8 [here](https://peps.python.org/pep-0008/#programming-recommendations))
* I think it would be better to have a single script with shared functions at the top and introduce additional functions for each of the examples and plots.
* Add references to  the original paper

murphyk:
@benlau6  Please update your code following gerdm's comments, then I will take a pass.

scripts/exact_1d_bayesian_filtering_utils.py
benlau6: Hi @gerdm, Do u mean that I should write the code like this?
```
def novel_density(
    rng_key, y, x_grid,
    x0_pdf, v_rvs: Callable, e_rvs: Callable,
    f: Callable, h: Callable, num_samples: int,
    K, kernel_variance,
):
```

gerdm: That looks much better! I would just modify the last line to be pep-8 compliant:

```
def novel_density(
    rng_key, y, x_grid,
    x0_pdf, v_rvs: Callable, e_rvs: Callable,
    f: Callable, h: Callable, num_samples: int,
    K, kernel_variance):
```

I would also suggest that if you're using type hints, be sure to use them consistently throughout the definitions of your functions.

benlau6:
@gerdm My scripts are revised. Please check. Note that I first committed changes into different files for your better diff comparison, and then committed to merge all those scripts into single file. Only small changes were done in this commit to make those examples into functions. Besides I deleted some helper functions which seems not helping at all. (point_mass_experimen and novel_experiment functions from utils.py)

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+
+def x0_rvs2(rng_key, shape):
+    return random.t(rng_key, df=3, shape=shape)
+
+
+# functions to get density
+pdf2 = functools.partial(jsp.stats.t.pdf, df=3)
+v_pdf2 = pdf2
+e_pdf2 = pdf2
+x0_pdf2 = pdf2
```
gerdm, line 385: If they are all the same, then it's redundant to define them over and over.

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+
+
+def student_t_random_walk_example(
```
gerdm, line 388: Following lines 382:385, it's better if you add `v_pdf2`, `e_pdf2`, and `x0_pdf2` as parameters to your function. Then, when calling the function, you could do something like: `student_t_random_walk_example (..., v_pdf2=pdf2, e_pdf2=pdf2, x0_pdf2=pdf2)`

scripts/exact_1d_bayesian_filtering_demo.py
gerdm: It would also be good to have more explicit names for your variables. Why is the `2` in the name?

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+
+    # plot the kth density
+    plot_density(
+        x_true, y, inv_measure_func3,
+        x_grid, p_pred, p_filter,
+        p_smooth, k=iter_, legend=True,
+        ax=None, title=f"Saturated measurements example densities at $x_{{{iter_}}}$",
+    )
+
+    plt.show()
```
gerdm, line 509: It might be better to have plt.show() after line `515`. The function should do the plotting and the user decides when to plot.

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+
+    # plot the kth density
+    plot_density(
+        x_true, y, inv_measure_func2,
+        x_grid, p_pred, p_filter,
+        p_smooth, k=iter_, legend=True,
+        ax=None, title=f"Student's t random walk example densities at $x_{{{iter_}}}$",
+    )
+
+    plt.show()
```
gerdm, line 426: See comment on L505

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+
+    # plot the kth density
+    plot_density(
+        x_true, y, inv_measure_func1,
+        x_grid, p_pred, p_filter,
+        p_smooth, k=iter_, legend=True,
+        ax=None, title=f"Particle filter example densities at $x_{{{iter_}}}$",
+    )
+
+    plt.show()
```
gerdm, line 348: See comment on L505

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+            linewidth=1.5,
+        )
+    fig.tight_layout()
+
+
+def experiment_setup(
+    rng_key, grid_minval, grid_maxval,
+    num_grid_points, x0_rvs, v_rvs,
+    e_rvs, f, h,
+    max_iter, plot_xy=False):
```
gerdm, line 245: Indent to distinguish from rest of code: see https://peps.python.org/pep-0008/#indentation

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+    v = v_rvs(rng_keys[0], shape=(max_iter + 1, num_samples))
+    e = e_rvs(rng_keys[1], shape=(max_iter + 1, num_samples))
+
+    for k in range(1, max_iter + 1):
+        x = inversion_sampling(
+            rng_keys[2],
+            x_grid,
+            p_filter[k - 1, :],
+            num_samples=num_samples,
+        )
```
gerdm, line 110: Have you tried [`jax.lax.scan`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html) on this code?

scripts/exact_1d_bayesian_filtering_demo.py
```diff
@@ -0,0 +1,515 @@
...
+    for k in range(max_iter - 1, -1, -1):
+        # p(xk, xk-1 | y(1:k-1))
+        px = x_pdf(k=k, x_new=X.T, x=X, v_pdf=v_pdf, f=f)
+        px = px * p_smooth[k + 1, :].T / p_pred[k + 1, :].T
+
+        p_smooth = p_smooth.at[k, :].set(jnp.sum(px, axis=1))  # marginalize
+        p_smooth = p_smooth.at[k, :].set(
+            p_smooth[k, :] * p_filter[k, :]
+        )  # multiply p(xk|yk)
+        p_smooth = p_smooth.at[k, :].set(p_smooth[k, :] / jnp.sum(p_smooth[k, :]))
```
gerdm, line 178: Try to use [jax.lax.scan](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html)

gerdm (changes requested):
Add some comments on style and possible increase in efficiently with jax.lax.scan

scripts/exact_1d_bayesian_filtering_demo.py
benlau6: Oh that sounds great. By the way I think that I need to use `student_t_random_walk_example (..., v_pdf=pdf2, e_pdf=pdf2, x0_pdf=pdf2)` instead, because `2` numbering is meaningless inside the example functions. On top of it, should I just declare internal functions `v_pdf`, `e_pdf`, `x0_pdf`, etc, in the example functions individually for encapsulation, or I should just use the existing practice, that define it outside and then pass it as argument?

e.g.
```
# state transition function
def state_trans_func2(x, v, k):
     return x / 2 + 25 * x / (1 + x**2) + 8 * jnp.cos(1.2 * (k + 1)) + v
...
def student_t_random_walk_example(
     ..., state_trans_func = state_trans_func2)

     ...
     state_trans_func()
```

vs

```
def student_t_random_walk_example(
     ...):

    def state_trans_func(x, v, k):
         return x / 2 + 25 * x / (1 + x**2) + 8 * jnp.cos(1.2 * (k + 1)) + v

    ...
    state_trans_func()
```

murphyk:
Sorry for the delay. I took a very quick look at the code and it seems correct. I refactored it into a colab (https://github.com/probml/probml-notebooks/blob/main/notebooks/discretized_ssm.ipynb) and ran it, and results seem similar to the paper - thanks!

murphyk:
@benlau6  Can you modify the colab code to plot every time step for the particle filtering example. Ideally we can plot true x(k), observed y(k), filtered mean E[x(k) | y(1:k)] and variance +- sqrt{Var[x(k)|y(1:k)} and smoothed  mean E[x(k) | y(1:T)] and variance +- sqrt{Var[x(k)|y(1:T)}, as a line plot. Also to capture multi-modality, we should plot the full posterior p(x(k) | y(1:k)) and p(x(k)|y(1:T)) as a heatmap (columns = time, rows = bins on x_grid).  I started this but ran out of time... Thanks!

benlau6:
@murphyk Sure, I am working on it. Besides, I would like to replace for loops with jax.lax.scan and a few changes as suggested by gerdm.

Issue #720, opened 2022-03-31
Create notebooks from scripts for book2: Chapter 18


patel-zeel:
# Instructions

* [Follow these guidelines](https://github.com/probml/pyprobml/blob/master/notebooks/README.md)

## Chapter 18: Gaussian processes
| Figure | Script | Notebook | PR | Author |
| :- | :- | :- | :- | :- |
| -- |-- |-- |-- |-- |

PR #727, opened 2022-04-03
Fig18.2 (Book2) add `gprDemoArd.ipynb`

patel-zeel:
Converted [gprDemoArd.py](https://github.com/probml/pyprobml/blob/master/scripts/gprDemoArd.py) to a notebook.

### Issue
#720 

### Figure Number

book2 - 18.2

### Figures

[gist](https://gist.github.com/patel-zeel/bd6d3601c0a77b7b20f436d287e8a96e)

## Checklist:

- [x] Tested on Google Colab.

Matplotlib tweaks for 3d figures are difficult and thus I added `bbox_extra='tight'`. This changes the figure size from desired but gets the work done (does not let the axis labels truncate). Thus, we need to manually fix the height (2 inches) for this figure in latex code.

cc @murphyk @nipunbatra


nipunbatra:
I have a few comments

- It is not clear what the colours mean. We would need to add a colorbar. 
- Would it be better to have the axis label "output" go from bottom to up on the axis? 
- How does a contour plot for the same data look like? Would the contour plot make it easier to explain the ARD functionality?
- It may perhaps be useful to show not only the function samples but perhaps also show fitting two GPs (GP 1 is w/o ARD, will have a poorer fit (marginal log-likelihood), GP 2 is w/ ARD). Such a dataset could be artificially generated.
- Which kernel are we using? Perhaps the caption can be edited to mention that as well?

patel-zeel:
> * It is not clear what the colours mean. We would need to add a colorbar.
> * Would it be better to have the axis label "output" go from bottom to up on the axis?

I have added a colorbar. I was unable to rotate the axis label because the `ax.set_zlabel` is not at all changing with the `rotation` argument.
image

> * How does a contour plot for the same data look like? Would the contour plot make it easier to explain the ARD functionality?

I think a 3d plot is somewhat more explainable but extremely difficult to modify. A contour plot looks like the following:
image

> * It may perhaps be useful to show not only the function samples but perhaps also show fitting two GPs (GP 1 is w/o ARD, will have a poorer fit (marginal log-likelihood), GP 2 is w/ ARD). Such a dataset could be artificially generated.

We may add this to the GP Regression section. What could be a good way to visualize the fit in 3d?

> * Which kernel are we using? Perhaps the caption can be edited to mention that as well?

I have updated the caption.

murphyk:
As discussed I think the plot with ARD "turned on" , so input x2 is ignore, is still unclear.

patel-zeel:
@murphyk I have changed the random seed and the updated figure looks like the following:

| Before | After |
| :-: | :-: |
| image | image |

murphyk:
Much better, thanks!

PR #728, opened 2022-04-04
Fig18.7 (Book2) add `gprDemoNoiseFreeAndNoisy.ipynb`

patel-zeel:
## Description

Added `gprDemoNoiseFreeAndNoisy.ipynb` for [gprDemoNoiseFree.py](https://github.com/probml/pyprobml/blob/master/scripts/gprDemoNoiseFree.py).

### Figure Number

book 2 - 18.7

### Figures

[gist](https://gist.github.com/patel-zeel/59c9983a8438eee648d077d58eebebf7)

### Issue 

#720 

## Checklist:

- [x] Tested on Google Colab.

I have added noisy regression plots as well (shown in the [above gist](https://gist.github.com/patel-zeel/59c9983a8438eee648d077d58eebebf7#newly-added-plot-noisy-regression)). Dr. @murphyk, please let me know if we should add the noisy regression figure. If yes, should it be in Section 18.3.2 or merged with Figure 18.7.

cc @nipunbatra


murphyk:
@patel-zeel  I suggest modifying fig 18.7 (17.7 in Apr 1 edn) to contain two rows: noise-free and noisy. Them modify text of sec 18.3.2 to refer to the latter figure.

patel-zeel:
@murphyk Given that the first figure will be the same for noise-free and noisy cases, would it be good to create three figures in a row like this? (I will open a new PR to include this change)

image

murphyk:
yes, good idea :)

Issue #745, opened 2022-04-19
translate remaining torch demos to JAX


murphyk:
- [x] [notebooks/layer_norm_torch](https://github.com/probml/probml-notebooks/blob/main/notebooks/layer_norm_torch.ipynb) - https://github.com/probml/probml-notebooks/pull/56, https://github.com/probml/probml-notebooks/pull/89
- [x] notebooks/text_preproc_torch - https://github.com/probml/probml-notebooks/pull/90
- [x] scripts/transposed_conv_torch
- [x] scripts/emnist_viz_pytorch

nalzok:
Note that `notebooks/layer_norm_torch` has already been translated in https://github.com/probml/probml-notebooks/pull/56

patel-zeel:
@nalzok I could not find `layer_norm_jax.ipynb` in `notebooks` folder, is it moved to another location?

nalzok:
Looks like it was accidentally removed in an unrelated PR: https://github.com/probml/probml-notebooks/pull/58. I will make a new PR now.

nalzok:
Hi @patel-zeel, it's done!

yuanx749:
Hi @patel-zeel , I converted `scripts/transposed_conv_torch` to jax.

PR #747, opened 2022-04-19
Add `emnist_viz_jax.ipynb`

nalzok:
### Colab link

https://colab.research.google.com/drive/1t8GBa00PFjDqcgyUI2xc8C5EgMjBQhVb?usp=sharing

### Issue 

https://github.com/probml/pyprobml/issues/745

## Checklist:

- [X] Tested on Google Colab.

Presumably, all scripts will be converted to notebooks in the end, so I just created a notebook for your convenience. This PR also showcases the power of the interactive visualization library [bokeh](https://github.com/bokeh/bokeh). I think it is more suitable for the Colab environment compared to matplotlib, and we should probably consider it in subsequent refactoring?

**EDIT:** Apparently there are 3 commits associated with this PR? I'm not sure why is that the case, but the [replace deprecated jax.tree_multimap with jax.tree_map](https://github.com/probml/pyprobml/pull/747/commits/268d0a6a0fa767fa75f37dd0b73c84f20ded341e) commit has already been merged and this PR only [adds one file](https://github.com/probml/pyprobml/pull/747/files), so I don't think there will be a merge conflict.


murphyk:
Thanks. Since this is a notebook I stored your code here:
 https://github.com/probml/probml-notebooks/blob/main/notebooks/emnist_viz_jax.ipynb
I will close this current issue since there seems to be a merge conflict.

PR #782, opened 2022-04-30
Wrap try accept: Fix static_import_check

karm-patel:
## Decryption
- Added `wrap_try_accept.py` Wrapped try..accept automatically on notebooks which was transferred from probml-notebooks/notebooks-d2l
- Removed the line `!pip install <package_name>`


karm-patel:
The above commit is regarding this issue: #785

Issue #783, opened 2022-04-30
Accelerate workflow by including SMOKE_RUN


patel-zeel:
After adding D2L notebooks, workflow execution time has increased from 1 hour to 2 hours 14 minutes. Ideally, the workflow should be as fast as possible so that we can know the failing components quickly. One of the solutions for this is to enable an environment variable called SMOKE_RUN for notebooks containing epoch-wise training with a logic that if SMOKE_RUN is in the environment we run only 1-2 epochs else we run the maximum epochs as mentioned in the notebook.

nalzok:
Since the tests have no dependency on each other, running all tests in parallel could be another viable option to accelerate the workflow. There are multiple ways to do this: generate a workflow for each notebook from some template, use [`pytest-split`](https://blog.jerrycodes.com/pytest-split-and-github-actions/), or use [`strategy.matrix`](https://hanxiao.io/2021/01/24/Speedup-CI-Workflow-in-Github-Actions-via-Strategy-Matrix/#building-job-matrix-for-parallelization).

The execution time can be further reduced if you add some [cache](https://github.com/actions/cache) so that the dependencies don't need to be installed each time.

patel-zeel:
@nalzok These all are excellent suggestions! Would you like to work on a PR to do this?

nalzok:
Sure, I’ll do some experiments to figure out which one works best and create a PR. It might take a few days though, since it’s the midterm week!

patel-zeel:
@nalzok I think the parallel workflow has improved the runtime a lot however, we can still improve it with SMOKE_RUN wherever possible. We can export the time taken to run a notebook along with the log on [this dashboard](https://github.com/probml/pyprobml/blob/workflow_testing_indicator/README.md). It will help us decide if a notebook needs to be fastened using SMOKE_RUN.

nalzok:
Yeah, it’s always better to have more transparency! Another thing I want to mention is that we can probably cache the datasets from TFDS and PyTorch-vision, since downloading mega bytes of data from the Internet is going to be slow.

patel-zeel:
@nalzok Yes, caching the datasets is a good idea. Is it possible in GitHub Actions? If not, we can try to maintain our own container with pre-downloaded datasets and a pre-configured Python version.

nalzok:
I think it’s possible with GitHub Actions. I will investigate into that later.

patel-zeel:
[...]

PR #792, opened 2022-05-01
Simple parallel CI workflow

nalzok:
## Description

Dead simple parallel CI workflow with 20 runners. Feel free to increase the number of concurrent jobs to 40 or something as long as it falls under the [usage limit](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits).

### Issue 

https://github.com/probml/pyprobml/issues/783

## Checklist:

- [X] Tested on Google Colab.

It's too risky to make use of [actions/cache@v3](https://github.com/actions/cache) without pinning the exact version of all packages, so this PR only implements parallel CI.

### DIFF
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -54,9 +54,16 @@ jobs:
         pytest -n auto -s --verbose tests/test_imports.py
 
   execute_notebooks:
+    strategy:
+      matrix:
+        runner_id: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
+    env:
+      PYPROBML_GA_RUNNER_ID: ${{ matrix.runner_id }}
+
     runs-on: ubuntu-latest
     # needs: [check_code_quality, static_import_check]
     container: texlive/texlive:TL2020-historic
+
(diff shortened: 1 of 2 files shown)


patel-zeel:
Nice trick using environment variable! Currently, all the other jobs are cancelled if one of them is failed. Shall we disable [fail-fast](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures)? 

I really like the idea of exetuting newer notebooks first because it saves lot of time of developers. Is it possible to implement that?

nalzok:
Hi @patel-zeel, it has been done! The notebooks are sorted by their modification time so that the most recently touched notebook is executed by job number 0, while the second most recently touched one is executed by job number 1, etc.

It took me a few commits to debug on the CI environment, but I have squashed the redundant commits with `git rebase` so that your commit log won't be polluted.

patel-zeel:
This looks great, @nalzok! I will soon create an automated dashboard to easily view the status and logs from the failing notebooks.

patel-zeel:
@nalzok I think it is difficult to get such workflows right in one commit so it is fine if there are multiple commits. This repo has only the "squash and merge" option so the commit log won't be polluted.

Issue #802, opened 2022-05-03
Explicit `try...except...` while importing packages


patel-zeel:
Currently, we import packages as follows:
```py
try:
    import tensorflow as tf
except:
    %pip install tensorflow
    import tensorflow as tf
```
This may hide some errors and thus a recommended way as pointed out by @nalzok [here](https://github.com/probml/pyprobml/pull/798#issuecomment-1116363421) is:
```py
try:
    import tensorflow as tf
except ModuleNotFoundError:
    %pip install tensorflow
    import tensorflow as tf
```

This needs to be applied systematically to all the notebooks.

PR #804, opened 2022-05-04
Make broad-except statements more specific

nalzok:
## Description

Address pylint warning [W0703 (broad-except)](https://vald-phoenix.github.io/pylint-errors/plerr/errors/exceptions/W0703.html). For example, the following pattern

```python
try:
    import tensorflow as tf
except:
    %pip install tensorflow
    import tensorflow as tf
```

is changed to

```python
try:
    import tensorflow as tf
except ModuleNotFoundError:
    %pip install tensorflow
    import tensorflow as tf
```

### Issue 

#802

The following scripts require additional care, since they intend to catch exceptions other than `ModuleNotFoundError`.

- `scripts/cholesky_demo.py`
- `scripts/lvae.py`

The following notebook is special since it contained a redundant `import tensorflow_probability as tfp` statement (which has been removed now)

- `notebooks/book1/13/mlp_1d_regression_hetero_tfp.ipynb`


patel-zeel:
@nalzok Sorry for not making it explicit earlier, but we are also deprecating the `scripts` folder, so we should make edits to their notebook versions from now onwards. The best place to search for a "script to notebook mapping" would be [this dashboard](https://github.com/probml/pyprobml/blob/workflow_testing_indicator/README.md) (also mentioned in the pinned issue). For example, the scripts suggested by you, live at the following locations as notebooks:

* [cholesky_demo](https://github.com/probml/pyprobml/tree/master/notebooks/book1/07/cholesky_demo.ipynb)
* [book2/20/LVAE](https://github.com/probml/pyprobml/tree/master/notebooks/book2/20/LVAE.ipynb)

nalzok:
Oops, somehow my regular expression missed these files. Let me fix them right away.

> Essentially, we can leave scripts in their current form for backward comparison.

OK, I will roll back the changes to `scripts/cholesky_demo.py` and `scripts/lvae.py`.

nalzok:
Hi @patel-zeel, I have fixed the problem, please review it again.

One thing to note is that `notebooks/book1/09/naive_bayes_mnist_jax.ipynb` contains a broad-except, but I am not sure which exception it is intended to catch, so I just put a `# TODO: ` comment there.

nalzok:
I did a `git merge master` to resolve the merge conflict and address the additional notebooks created since I started working on this branch. The following notebooks require special care

- `notebooks/misc/clip_imagenette_make_dataset_pytorch.ipynb`
- `notebooks/misc/clip_imagenette_demo.ipynb`
- `notebooks/misc/clip_make_dataset_tpu_jax.ipynb`

They contain curious patterns like the following, are they trying to a `ConnectionResetError` or something network-related?

```python
import tensorflow_datasets as tfds

try:
    data, info = tfds.load("imagenette/160px-v2", with_info=True, as_supervised=True)
except:
    data, info = tfds.load("imagenette/160px-v2", with_info=True, as_supervised=True)
    train_data, valid_data = data["train"], data["validation"]"
```

murphyk:
The following idiom  is a hack we created to compensate for the fact that tfds often times out the fits time you call it.
There may be a better way...

```

try:
    data, info = tfds.load("imagenette/160px-v2", with_info=True, as_supervised=True)
except:
    data, info = tfds.load("imagenette/160px-v2", with_info=True, as_supervised=True)
    train_data, valid_data = data["train"], data["validation"]"
```

patel-zeel:
> One thing to note is that `notebooks/book1/09/naive_bayes_mnist_jax.ipynb` contains a broad-except, but I am not sure which exception it is intended to catch, so I just put a `# TODO: ` comment there.

I see the following structure at # TODO:
```py
try:
    img = np.array(img)
except Exception:
    pass
```
Do you know in which case the above exception would occur? If code is trying to convert a `non-ndarray` object to `ndarray`, why should we `pass` if it fails to do so?

nalzok:
> Do you know in which case the above exception would occur? If code is trying to convert a `non-ndarray` object to `ndarray`, why should we `pass` if it fails to do so?

I just did some investigation, and it turns out that this snippet was proposed by @murphyk in [this comment](https://github.com/probml/probml-notebooks/pull/52#issuecomment-1082477098), which references [`torch.py` in the d2l repository](https://github.com/d2l-ai/d2l-en/blob/master/d2l/torch.py), which defines `d2l.numpy` as

```python
numpy = lambda x, *args, **kwargs: x.detach().numpy(*args, **kwargs)
```

I guess the except clause deals with the case where `img` is already a numpy array, which does not have a `detach` method. The try-catch statement is redundant for `np.array`, so let's get rid of it.

Issue #739, opened 2022-04-14
[Dashboard] [book1] [Chap: 1 to 10] Move from python scripts to notebooks


patel-zeel:
## Instructions

* Follow [the contributing guidelines](https://github.com/probml/pyprobml/blob/master/CONTRIBUTING.md) and specific instructions given over [here](https://github.com/probml/pyprobml/blob/master/notebooks/README.md).

## Dashboard

<details open>
<summary>Chapter: 01_introduction</summary>

|   fig | notebook                                                                                           | workflow                                                                                                                                                                                                                                                                                             | fine-tune      | latexify       | jaxify         | interactive    |
|------:|:---------------------------------------------------------------------------------------------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:---------------|:---------------|:---------------|:---------------|
|  1.3  | [iris_plot.ipynb](https://github.com/probml/pyprobml/01/iris_plot.ipynb)                           | ('<img width="20" alt="image" src=https://raw.githubusercontent.com/probml/pyprobml/workflow_testing_indicator/notebooks/book1/01/iris_plot.png>', '[log](https://raw.githubusercontent.com/probml/pyprobml/workflow_testing_indicator/notebooks/book1/01/iris_plot.log)')                           | (&#10060;, PR) | (&#10060;, PR) | (&#10060;, PR) | (&#10060;, PR) |
|  1.5  | [linreg_residuals_plot.ipynb](https://github.com/probml/pyprobml/01/linreg_residuals_plot.ipynb)   | ('<img width="20" alt="image" src=https://raw.githubusercontent.com/probml/pyprobml/workflow_testing_indicator/notebooks/book1/01/linreg_residuals_plot.png>', '[log](https://raw.githubusercontent.com/probml/pyprobml/workflow_testing_indicator/notebooks/book1/01/linreg_residuals_plot.log)')   | (&#10060;, PR) | (&#10060;, PR) | (&#10060;, PR) | (&#10060;, PR) |
[...]

PR #821, opened 2022-05-12
Updated iris_logreg

nitish1295:
## Description

Updated iris_logreg.ipynb for [iris_logreg.ipynb](https://github.com/probml/pyprobml/blob/master/notebooks/book1/02/iris_logreg.ipynb). 

Kindly note that this notebook contains 2 figures 2.11 and 2.13 from Book1, figure 2.11 is already updated, and this PR contains changes for Fig 2.13 only

### Figure Number
Book 1 - 2.13

### Figures

[gist](https://gist.github.com/nitish1295/5026076cf60b1fe8e89ca8c3ff2cf625)

### Issue
https://github.com/probml/pyprobml/issues/739

## Steps

- [x] Changed the name of the variables/functions to PEP 8 style
- [x] Added code to generate figure 2.13 in the book
- [x] Converted code to JAX

## Checklist:

- [x] Tested on Google Colab.


murphyk:
The color scheme needs to be changed in the way Nipun suggested.

nipunbatra:
Showing three classes with probabilities is not as intuitive as for two classes. 

I think we have two options

## Option 1
Show the class probabilities for each class in a separate figure


image

image

image


## Option 2
Show the class probabilities for all the classes in a single figure

image


@murphyk please let us know which you'd prefer.

murphyk:
I prefer option 2.

nitish1295:
Dr @murphyk I have updated the notebook based on the suggestions above. Kindly refer to the updated [gist](https://gist.github.com/nitish1295/5026076cf60b1fe8e89ca8c3ff2cf625)

murphyk:
The plot looks good. However the probabilities are very saturated. I wonder if you can reduce the "overconfidence" by adding some L2 regularization, and/or reducing the number of samples per class? This will make the plot more "interesting".

nitish1295:
Dr @murphyk, kindly let us know your thoughts about the following figure

image

murphyk:
Let me merge the current version, we may tweak this again later.

PR #846, opened 2022-05-18
Add autogenerated contributors' list

patel-zeel:
## Steps

* [A script](https://github.com/patel-zeel/pyprobml/blob/contributors/internal/contributors/contributors.py) is added to autogenerate the contributors based on [this configuration](https://github.com/patel-zeel/pyprobml/blob/contributors/internal/contributors/contributors.toml).
* Modified README.md file with updated list of contributors


internal/contributors/contributors.py
```diff
@@ -0,0 +1,55 @@
+import pandas as pd
+import toml
+
+## Config
+with open("internal/contributors/contributors.toml") as f:
```
murphyk, line 5: Why do we need to factor out the config file? Maybe just hard-code the values in this script, for simplicity?
(One less thing to change.)

internal/contributors/contributors.py
```diff
@@ -0,0 +1,55 @@
...
+pull_requests = pd.concat(pr_df_list)
+
+## Get count of PRs per contributor
+
+pull_requests["login"] = pull_requests["user"].apply(lambda x: x["login"])
+contributor_pr = pull_requests.groupby("login").agg({"url": len}).sort_values(by="url", ascending=False)
+contributor_pr.rename(columns={"url": "Number of PRs"}, inplace=True)
+
+# Filtering
+atleast_2_pr = contributor_pr[contributor_pr["Number of PRs"] >= config["min_PRs"]]
```
murphyk, line 33: Rename to `atleast_n_pr` since n may not be 2 :)

internal/contributors/contributors.py
```diff
@@ -0,0 +1,55 @@
...
+    page_no += 1
+pull_requests = pd.concat(pr_df_list)
+
+## Get count of PRs per contributor
+
+pull_requests["login"] = pull_requests["user"].apply(lambda x: x["login"])
+contributor_pr = pull_requests.groupby("login").agg({"url": len}).sort_values(by="url", ascending=False)
+contributor_pr.rename(columns={"url": "Number of PRs"}, inplace=True)
+
+# Filtering
```
murphyk, line 32: Before filtering, it might be worth printing the number of PRs and commits, to make sure we are not being too aggressive.

murphyk (changes requested):

internal/contributors/contributors.py
patel-zeel: Right, deleting the `.toml` file.

internal/contributors/contributors.py
patel-zeel: Done.

internal/contributors/contributors.py
patel-zeel: Does that mean if we are leaving out a few contributors with a single PR or a single commit, we should include them? In current settings, we are considering all contributors with at least two PRs **OR** at least two commits.

patel-zeel: Done

Issue #855, opened 2022-05-20
remove duplicated notebooks


murphyk:
Many notebooks in https://github.com/probml/pyprobml/tree/master/notebooks/misc have been copied to notebooks/book1 or notebooks/book2, sometimes with a changed name (eg https://github.com/probml/pyprobml/blob/master/notebooks/misc/bnn_mnist_sgld_whitejax.ipynb -> https://github.com/probml/pyprobml/blob/master/notebooks/book2/18/bnn_mnist_sgld.ipynb). 
Please move all such duplicates out of notebooks/misc and put in https://github.com/probml/pyprobml/tree/master/deprecated.

The final result should be that notebooks/misc just contains "orphans", ie notebooks that are not referred to anywhere in either book. It will be easier to identify a home for these orphans once there are fewer of them :)

karm-patel:
Dr. @murphyk, I have moved 49 notebooks having the same names. Still, there are 130 notebooks left in `notebooks/misc`.
I'm still figuring out the way to find out duplicate notebooks which do not have the same names.

PR #858, opened 2022-05-21
Moved duplicated notebooks  `/deprecated`

karm-patel:
## Description
I have moved 49 duplicated notebooks from `notebooks/misc` to `deprecated/`. These notebooks have the **same names**.

## Issue
#855


murphyk:
Thanks @karm-patel . There are now 125 files in notebooks/misc. Some of these are referred to in the supplementary material sections of book 1. For example,
https://github.com/probml/pyprobml/tree/master/notebooks/book1/01#supplementary-material
(intro to pandas) links to 
https://colab.research.google.com/github/probml/probml-notebooks/blob/master/notebooks/pandas_intro.ipynb.
Can you please identify all the notebooks which are pointed to in the markdown files and move them from notebooks/misc
 to the relevant chapter directory (eg notebooks/book1/01). (And update the readme links to point to the new location, a not to  probml-notebooks). I think this will get rid of ~10 orphans :)

Thanks

PR #892, opened 2022-06-03
create plot of dp mixture simulation (fig 30.4 in book2); and plot of…

xinglong-li:
… clustering analysis using dp (fig 30.6 in book2)

## Checklist

- [ ] Performed a self-review of the code
- [ ] Tested on Google Colab.

### DIFF
--- a/internal/book2/create_figure_cluster_dp_mix.py
+++ b/internal/book2/create_figure_cluster_dp_mix.py
@@ -1,0 +1,180 @@
+import jax
+import jax.numpy as jnp
+import numpy as np
+from jax.scipy.special import gammaln
+from jax.numpy.linalg import slogdet, solve
+from jax import random
+from collections import namedtuple
+
+                
+@jax.jit
+def log_p_of_multi_t(x, nu, mu, Sigma):
+    """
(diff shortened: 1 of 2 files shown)


murphyk:
- The https://github.com/probml/pyprobml/tree/master/internal directory is used for internal book-keeping stuff, like processing the latex, and is not the right place to store user-facing demos.
- You write `from NIW import NormalInverseWishart` but it is not clear what the NIW package is.
- For generic utility / library code  (eg the multivariate T functions, and `dp_cluster` function), make a new file `dp_mixgauss_utils.py` and store it in https://github.com/probml/probml-utils/tree/main/probml_utils.
- Each figure for the book should have a corresponding ipynb notebook. So you should create `dp_mixgauss_sample.ipynb` for 30.4 and `dp_mixgauss_fit.ipynb` for 30.6. Store these in https://github.com/probml/pyprobml/tree/master/notebooks/book2/30.  
- You can import your utilities file into the notebook following the idiom in https://github.com/probml/pyprobml/blob/master/notebooks/book2/30/dpm_sample_demo.ipynb.

Issue #862, opened 2022-05-23
Refactoring Pending tasks


karm-patel:
I am putting a list of pending tasks in refactoring the repo. Feel free to add more.
1. - [x] Add `README.md` file in `notebooks/book2` which contains chapter name.
2. - [ ] Remove `_jax` from the notebook name.
3. - [ ] Postprocess the notebook of book2 i.e remove `!pip install`
4. - [ ] Resolve trivial workflow errors of the book2    
5. - [ ] Resolve non-trivial workflow errors of book1 (mostly d2l notebooks are remaining)
6. - [x] Figure out a way to update latexify/jaxify status in the dashboard.
7. - [ ] Add pointer to figure (png and pdf) both in chapterwise README.md in book1
8. - [x] Figure out the height of each figure programmatically, so while doing `latexify` one can easily refer to that mapping without manual checking in latex source code. 
9. - [ ] Move unnecessary figures from `bookv2/figures/` to `bookv2/extra_figures`
10. - [ ] Remove try..except from `probml_utils` and add those modules in req.txt (in probml-utils) or req-exra.txt ((in pyprobml)
11. - [x] Remove `_latexified` from fig_name while calling `savefig()` instead add it automatically.
12. - [ ] Create a common .md file containing fig_no, notebook name, so we can get rid of firestore.

patel-zeel:
@nalzok can you take care of the workflow errors in d2l notebooks as and when you refactor them?

nalzok:
Sure, I will take care of the D2L notebooks!

murphyk:
Hi @karm-patel . I think this list needs udpating :)

karm-patel:
Let me close this, I'm no more using this issue to track pending tasks.

PR #895, opened 2022-06-04
Remove `_latexified` from figure names in notebooks

karm-patel:
## Description

Removed suffix `_latexified` from all figure names in all notebooks. Since `savefig()` will add it automatically when `LATEXIFY` is enabled. we did this to avoid spelling mistakes from users.

## Issue
Task 11 of #862


murphyk:
good idea! :)

PR #899, opened 2022-06-06
Book 2 | Fig 3.7

VibhutiBansal-11:
## Description
Latexified, jaxified the code.

### Figure Number
Book 2 Fig 3.7 

### Figures
|Before|After|
|---|---|
|before|image

### Issue 

#739 

## Steps

- [x] Latexified the code
- [x] Jaxified the code
- [x] Added interactive demo for eta and random seed
## Checklist:

- [x] Tested on Google Colab.

- using font as, font_size = 6 if is_latexify_enabled() else None due to small figure size & overlapping legend
- renamed the file from lkj_numpyro to lkj_1d as tensorflow probabilty is being used in the code instead of numpyro
cc Dr. @murphyk , Prof. @nipunbatra


karm-patel:
Dr. @murphyk, this file is renamed to `lkj_1d.ipynb` from `lkj_numpyro.ipynb`, since the code is not using numpyro now. So what to do with the old notebook? we can have the following alternatives.
1. Delete the old notebook and update the caption to link the new notebook.
2. We can update the caption which can point to the latest notebook but we can keep the old notebook and add a pointer link to another notebook in both.  

I'm inclined towards option 2, so in the future, we can compare the results.

murphyk:
> Dr. @murphyk, this file is renamed to `lkj_1d.ipynb` from `lkj_numpyro.ipynb`, since the code is not using numpyro now. So what to do with the old notebook? we can have the following alternatives.
> 
> 1. Delete the old notebook and update the caption to link the new notebook.
> 2. We can update the caption which can point to the latest notebook but we can keep the old notebook and add a pointer link to another notebook in both.
> 
> I'm inclined towards option 2, so in the future, we can compare the results.

Yes, lets do option 2.

Issue #890, opened 2022-06-01
Refactor notebooks | Chapter: Markov Chain Monte Carlo inference | Book2


karm-patel:
This issue is created to cover all latexify/jaxify and new notebook creations.

**To dos**
- [x] 11.14 to 11.17: Change notebook from `mcmc_traceplots_unigauss_numpyro` to `mcmc_traceplots_unigauss`
- [x] 11.18: Add new notebook, Update caption as suggested [here](https://github.com/probml/pyprobml/pull/913#issuecomment-1158893508)
- [x] 11.20: Remove `funnel_numpyro.ipynb` & keep only `neals_funnel.ipynb`

PR #901, opened 2022-06-07
plot for fig 30.4 and 30.6

xinglong-li:
add   dp_mixgauss_forward.ipynb   to reproduce fig30.4
add   dp_mixgauss_cluster.ipynb   to reprocuce fig30.6

PR #913, opened 2022-06-17
Added `rhat_slow_mixing_chains.ipynb` for Fig 11.18 | book2

karm-patel:
## Description

I have added a notebook for figure 11.18 which was screenshot previously. I have considered following points in notebook: 

1. Current figure does not show the target distribution from where these samples are taken. I have also checked the references ([paper](https://arxiv.org/abs/1903.08008) & [book](http://www.stat.columbia.edu/~gelman/book/) - Fig 11.3 by Dr. Vehtari) given in caption, but they also didn't mention target distribution. Hence I took bimodal normal distribution as shown in the first subplot.
2. In the book2, section-11.6.2.2, Dr. Kevin mentioned that "_Although the above definition can detect non-convergence of the kind shown in Figure 11.18(a), it will fail on non-stationary chains of the kind shown in Figure 11.18(b)._" But in the current figure, we can not visually see if **old $\hat{R}$** is failing in the 2nd subplot or not (since the author didn't show the numbers). So I have calculated this old $\hat{R}$ for both situations.  
which would help to understand the insight- "we are getting $\hat{R}=1.003$ despite the chain is not converged yet."  

Dr. @murphyk, please review the above changes. 
## Question
1. I am confused about how to mention R_hat (which is calculated as per the old definition) in the title? does the current version look good or do we need like - old $\hat{R} = 1.003$ 
- One alternative could be, keep this version and mention in caption clearly that  $\hat{R}$ is calculated as per old definition using equation x.y 

### Figure Number
Book2: Fig 11.18

### Figures
Before PR:
image

After PR:
image

### Issue 
#890 

cc: Dr. @murphyk, Prof. @nipunbatra


karm-patel:
Dr @murphyk, First two checks are failing due to this commit: https://github.com/probml/pyprobml/pull/901

murphyk:
I suggest renaming `calculate_arviz_rhat` to `calculate_split_R_hat`, and add a comment explaining how the latter is defined.
In the figures, you should print both `nonsplit-Rhat` and `split-Rhat` in the title.
Also in the figure caption (latex) you should explain that the samples in both cases (b and c) are generated using HMC, but for (b) you use 
```
step_size = 0.05
inverse_mass_matrix = jnp.array([0.8])
integration_step_size = 13
```
and (c) you use
```
step_size = 0.1
inverse_mass_matrix = jnp.array([0.1])
integration_step_size = 13
```

PR #918, opened 2022-06-19
Latexified/Jaxified Fig 11.20 `neals_funnel.ipynb`

karm-patel:
## Description

Changes:
1. Converted `funnel_numpyro.ipynb` to blackjax
2. I have merged the **old code** for 11.20(a) & 11.20(b, c) into a single notebook - `funnel_numpyro.ipynb`. we are keeping this for comparison purposes.
3. I have merged the **updated code** for 11.20(a) & 11.20(b, c) into a single notebook - `neals_funnel.ipynb`. So we need to change caption as well. 

I have the following confusion in related section 11.6.4:
1. In Eq. 11.123 & 11.126, I think we should replace `v` to `v/2` (Image1 & Image2). I have written code considering `v/2`. ([reference](https://num.pyro.ai/en/latest/examples/funnel.html))
**Image1:**
image

**Image2:**
image

2. The term `variance v` looks confusing to me (Image3), because if we are doing $exp(v/2)$ then shouldn't $v$ looks like log (variance)?

**Image3:**
image

### Figure Number
Fig 11.20

### Figures
**Before PR:**
image

**After PR:**
image

### Issue 
#890 

cc: Dr. @murphyk, Prof. @nipunbatra


murphyk:
Yes, v=log variance. Fine to use v/2 to match numpyro demo.

PR #932, opened 2022-06-25
Added  Common readme file

karm-patel:
## Description

- Added [common_readme.md](https://github.com/karm-patel/pyprobml/blob/common_readme/internal/common_readme.md) which contains notebook_name, colab link and github link

I am enlisting advantages of having this file just for recap:
1. Now we don't need firestore redirection, so adding a new notebook would be easy. 
2. We don't need duplicate notebooks in separate folders which point to the root notebook.
   - Currently, we can remove book2 duplicate notebooks, but not book1 to maintain backward compatibility.
3. We don't need dummy notebooks like `fig_10_9.ipynb` anymore.
   - Again we can remove `fig_x_y.ipynb` from book2, but not book1

- There were some notebooks with identical content or either of them latexified and other is not. I've updated all such notebooks such that we will have only a single root notebook and other notebooks would have pointers to this root notebook.


karm-patel:
Dr. @murphyk, in which directory you will suggest putting `common_readme.md` >

PR #933, opened 2022-06-26
create: v2.16-bayesian_neural_network

gerdm:
## Description

Example of Bayesian neural network using flax + blackjax

### Figures

<p align="left">
    
    
</p>


nipunbatra:
Hi @gerdm It may be useful to also plot the contours for the variance. It'll show that the MLP does not do a good job for OOD data. Colour bar on the side may also help interpret the colours better (but perhaps that depends on how detailed we want to be)

gerdm:
Hi @nipunbatra,

I like the idea of having a contour plot with the posterior variance and a colorbar. For the posterior mean, I don't think we need a colorbar. I think the contrast in colours are pretty explanatory.

murphyk:
This is great. But can you also add a mini-batch SG-MCMC version to show what happens when N gets larger?

murphyk:
The filename is a bit generic. (We will be adding other examples of "bayesian neural networks" soon)
Can you rename to `bnn_mlp_2d_hmc`?

karm-patel:
Hi @gerdm, Can you please run `black --verbose --line-length=120 <path/to/bayesian_neural_network.ipynb>` , in order to pass first workflow?

image

PR #940, opened 2022-06-26
Removed suffix like `_pymc`, `_blackjax`, etc from book2 notebooks

karm-patel:
## Description

1. Removed following suffixes from 23 notebooks.
```py
["tfp","blackjax","jax","pyro","numpyro","pymc3","pymc","sklearn","flax"]
```
2. I've updated the notebook name in book2 `.tex` files, please see this PR:https://github.com/probml/bookv2/pull/80 
3. I've also updated the other notebooks wherever these old names were referred.

**Note:** **This change will break 23 URLs in the current notebook.**


murphyk:
there is some kind of error: https://github.com/probml/pyprobml/runs/7063022379?check_suite_focus=true

karm-patel:
> there is some kind of error: https://github.com/probml/pyprobml/runs/7063022379?check_suite_focus=true

Dr. @murphyk, Yes right, I'm resolving these errors in other threads, currently, I'm merging this PR so we can test dead URLs in bookv2's workflow.

PR #1078, opened 2022-07-27
rewrite dp_mixgauss_cluster notebook

xinglong-li:


murphyk:
It would be better to replace
```
samples_set[f"gmm_{n_clusters}_data_{str(m_precision).split('.')[0]}p{str(m_precision).split('.')[1]}"] = gmm_samples
```
with something like this
```
def replace_period(m):
  return str(m).split('.')[0]}p.split('.')[1]

fname = "gmm_{n_clusters}_data_{replace_period(m_precision)}}"
samples_set[fname] = gmm_samples
```

xinglong-li:
Thanks, I'll change it in my future code!

PR #1079, opened 2022-07-27
Moons

AnandShegde:
## Description
Created notebook comparing the following techniques on Moons dataset.
* MLP
* Deep Ensemble
* MC Dropout
* Bootstrap
* NUTS
* GP
* SNGP
* VI

### Figures
image

## Checklist:

- [x] Tested on Google Colab.


murphyk:
LGTM. One concern is that SNGP uses `DeepResNet(tf.keras.Model)` which is a different model to `cmlp`, so we're conflating two things here - the architecture and the estimation method. But changing this may be hard, so I will merge for now.

PR #1083, opened 2022-07-27
Add GHKF plots to ekf_vs_ukf.ipynb

petergchang:
## Description

Added GHKF plots to `ekf_vs_ukf.ipynb`.

### Figure Number
book2 - 8.12

### Figures
ghkf_source ghkf_target

### Issue 
#1082 

## Checklist

- [x] Tested on Google Colab.


murphyk:
Can you please make helper functions 
```
(m,S,C) = ukf.transform(non_linear_transform, mean_pre, cov_pre)
(m,S,C) = ghkf.transform(non_linear_transform, mean_pre, cov_pre)
```
that wraps up some of the low level logic (I think you can reuse https://github.com/probml/ssm-jax/blob/main/ssm_jax/ggssm/containers.py#L61).
Something like this:

Screen Shot 2022-07-27 at 12 39 49 PM

I realize that you may also want to return the transformed sigma points,  for plotting, so these
could be optional return args.

PR #1106, opened 2022-09-11
Minor bug in computing the energy of ising demo.

lkhphuc:
The contribution of the lower y pixel is assigned to an unused variable a instead of to wi.

### Figure Number

book2 Figure 4.17

### Figures

| Before PR | After PR |
| :-: | :-: |
|image|image
|image|image
|image|image|
## Checklist

- [x] Tested on Google Colab.

At J=0.4 the updated PR has more "clumpiness" than original,  but the phase transition for this model is ~0.44 per Eq. 4.79

### DIFF
--- a/notebooks/book2/04/gibbs_demo_ising.ipynb
+++ b/notebooks/book2/04/gibbs_demo_ising.ipynb
@@ -53,7 +53,7 @@
     "\n",
     "def energy(ix, iy, X, J):\n",
     "    wi = 0\n",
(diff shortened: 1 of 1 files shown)


Neoanarika:
I checked the changes. I agree with there was a bug and that the PR corrected the mistake. Thanks on the Gibbs Ising code.

Neoanarika (commented):
Yup, this is an unused variable and should be assign to wi

Neoanarika (commented):
Thanks

